PR #5, opened 2015-01-03
UI corrections

dcell:
There was a tab in the line that positions the -h flag in the menu, causing it to be slightly displaced when you look at the help menu. There was also a discrepancy between the data source "GooglePlus" and "googleplus", the all lower case one being used in the actual programming and the capitalized version in the help menu, So I just changed the one in the help menu to be all lower case.

### DIFF
--- a/theHarvester.py
+++ b/theHarvester.py
@@ -39,7 +39,7 @@ def usage():
                         people123
                         jigsaw
                         twitter
-                        GooglePlus
+                        googleplus
                         all\n"""
     print "       -s: Start in result number X (default 0)"
     print "       -v: Verify host name via dns resolution and search for virtual hosts"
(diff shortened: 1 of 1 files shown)


leebaird:
I have already corrected this and much more with my pull request.

dcell:
I saw that.

But to be perfectly honest, I got the impression your pull request was more under the pretense of changing the program, since you made changes to almost every file, rewriting the majority of what you changed. Much of it completely unrecognizable. Not to mention a large amount of what you considered corrections, were just rewrites on already correct code to change it to a way you prefer it. Thus largely subjective. 

Mine is more selective to the UI, and doesn't pull in a borderline complete rewrite of everything along with it, which is why I submitted it anyway.

laramies:
Thanks Joshua for the fix.

PR #7, opened 2015-01-12
Adding support for private search engines

gilnajera:
Https support for DuckDuckGo, ixquick and Disconnect.me


NotoriousRebel:
@gilnajera please fix conflicts and test results. I tested with duckduckgo and it found nothing.

PR #8, opened 2015-01-16
Improve usage() when installed in /usr/bin

CameronNemo:
This makes this patch (from Kali Linux) obsolete:

http://git.kali.org/gitweb/?p=packages/theharvester.git;a=blob;f=debian/patches/edit-usage-output;h=53e4bf761725411ea64e8469f0edd97c6438c084;hb=HEAD

### DIFF
--- a/theHarvester.py
+++ b/theHarvester.py
@@ -3,6 +3,7 @@
 import string
 import httplib
 import sys
+import os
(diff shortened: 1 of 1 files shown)


CameronNemo:
Ping?

PR #10, opened 2015-02-23
Removed unneeded debugging lines

dcell:
It seems like these lines might have been left over when you were debugging it, as they print unparsed data while running the program.


laramies:
Thanks!

PR #11, opened 2015-03-05
Fix EXTRAword.com matching when word is 'word.com'

rtkpmcalpine:
This change makes myparser.emails(...) stricter so that, given a word
(eg: 'word.com') it does not match domains which end with that word
(i.e. some-other-word.com). Subdomains continue to be matched (eg:
'subdomain.word.com').

Also add unit test for emails(...)


laramies:
Thanks, works fine :)

PR #12, opened 2015-03-25
added new search engines

tatanus:
I added two new search engines (baidu and yahoo)
I also updated theHarvester.py and discovery/__init__.py accordingly.
I also fixed a coupy typoes in theHarvester.py whgile I was at it.


darker0n:
Cool :+1:

laramies:
Thanks Tatanus! :)  I will change later Httplib by Requests as I did with Google and other search plugins.

PR #22, opened 2015-06-14
Small email improvement

Th4nat0s:
Hello,  a small contribution.

change the regex to avoid invalid results like "@mydomain.com" without private part.
add "| uniq | sort" to email list to avoid duplicates in results.
add -m switch for people who only needs emails :) ...

Still need a "lower(email)" somewhere....

Bye.

### DIFF
--- a/myparser.py
+++ b/myparser.py
@@ -33,7 +33,9 @@ class parser:
     def emails(self):
         self.genericClean()
         reg_emails = re.compile(
-            '[a-zA-Z0-9.-_]*' +
+            # Local part is required, charset is flexible
+            # https://tools.ietf.org/html/rfc6531
+            '[a-zA-Z0-9.\-_+#~!$&\'()*,;=:]+' +
             '@' +
             '[a-zA-Z0-9.-]*' +
             self.word)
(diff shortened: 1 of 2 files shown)


laramies:
Thanks for this, I will look for this in detail soon.

tambry:
"soon"?

laramies:
Hi, I pushed myparser.py changes in Master (removed * and () ) as it was producing lot's of dirty results in my tests.  Regarding the email only, I can add a switch for the output printing only, as the host extraction it virtually doesn't take any time.  Will that do it?   (I couldn't pull this changes into master, thus why I made them myself)

Th4nat0s:
I see conflict do you need another merge request ?

laramies:
Please do a merge request for the e-mail only option, please. :)

Th4nat0s:
Do you mean only the -m "option".. ?? but i'm really interested  (WANT IT in fact :)… ) by the | sort | uniq.. may i implement it in "-m" ... ?

laramies:
If by sort | uniq you mean    `all_emails = sorted(set(all_emails))`  I already added that.  I was referring to `-m`

leebaird:
First, second, and fourth requests have been added.

laramies:
Closing as most request are implemented, except to the option to retrieve the emails only, which I dont think is needed, you can export to XML and grab the emails.

PR #28, opened 2015-07-10
Very minor aesthetic code updates.

initbar:
Hello,

I've only touched on those which could be expressed in different styles (i.e. `if shodan == True` with `if shodan`, or `if all_hosts == []` with `if not all_hosts`).

I made the last `except`, however, as `finally` in order to act as a clean-up action. If `sys.exit()` is placed as another `except`, that would mean `sys.exit()` would only be called only if the `except` logic was reached.


initbar:
Closing to keep the code style of the author.

PR #33, opened 2015-12-02
Fixed the order of two lines in the command line example

digininja:
The command line example had two lines switched around so fixed them.

Also fixed a couple of PEP 8 complaints.


laramies:
Thanks!

PR #37, opened 2016-02-13
xml-output with shodan and more

surfuga:
1-xml-output: in the host, vhost I've added a subfield (ip and hostname)
2-xml-output: adding shodan results
3-adding sorted(set()) to all_emails, all_hosts, vhosts
4-using full in the xml-output for the host


laramies:
Thanks, pushed to Master. :)

PR #39, opened 2016-03-05
fix incorrect file closing

initbar:
After writing to save emails, hosts, and vhosts to a file, the `file` object never properly calls `.close()` (but invoke as an attribute `.close`). This will always trigger `Exception` thereby never exporting the contents in the first place.


laramies:
Thanks initbar :)

PR #67, opened 2017-09-04
Filename now strips after the last full stop, not the first one

m-nerek:
For the current implementation,  feeding a filename to save such as 

`-f /foo/website.com/results.extension`
results in the file being saved as 
`/foo/website.xml`

This quick fix changes that to have the expected result of saving it as:
`/foo/website.com/results.xml`.
Utilizing `x.rsplit`, instead of `x.split`.

### DIFF
--- a/theHarvester.py
+++ b/theHarvester.py
@@ -400,7 +400,7 @@ def start(argv):
             print e
             print "Error creating the file"
         try:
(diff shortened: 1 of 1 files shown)


laramies:
Thanks, fixed in last release!

PR #69, opened 2017-10-08
Added Google Certificate transparency report support

wez3:
Queries the Google Certificate transparency report and lists the discovered domains in the results.


laramies:
I missed many notifications in the past, I merged it now! thanks for your collaboration!

PR #81, opened 2018-03-13
Shodan update - to use built-in modules

sarlacii:
Hi there

Not sure about my changes to main file ("theHarvester.py") as it's pretty subjective depending on what you want out. I opted for a shorter list, with just service exposed (version and timestamp).

Main change is to the shodansearch.py file, to use built-in python2 module, instead of the API in the shodan directory.


NotoriousRebel:
Hi @sarlacii 

Thank you for your contributions, and apologies for getting to this so late.
The issue you have raised has been resolved and we appreciate you bringing it up.

As things move forward and theHarvester is elevated to another level thanks to the fantastic community, pull requests will be monitored and responded to much faster.

Thank you for your contribution.

PR #85, opened 2018-07-07
All classes for search engines should ideally have the same interface.

MarioVilas:


MarioVilas:
Never mind, there are many more inconsistencies anyway.

PR #90, opened 2018-09-08
removed some imports, replaced httplib with requests, cleaned up exceptions and print statements

yaleman:
* httplib isn't used in theHarvester.py, so importing it is unnecessary
* shodan searches for "empty" as an IP address don't need to be done, so I added a check to skip them
* replaced httplib in discovery with requests
* cleaned up some exceptions, added messages
* changed print usage from statements to functions

### DIFF
--- a/discovery/DNS/Base.py
+++ b/discovery/DNS/Base.py
@@ -33,7 +33,7 @@ def ParseResolvConf(resolv_path):
     try:
         lines = open(resolv_path).readlines()
     except:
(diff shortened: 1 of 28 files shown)


laramies:
Hi @yaleman,  we just released a new version that replaces httplib and is python3 compatible. II will close this PR as it wont apply now. If there is anything that still can be improved please submit a new PR. Thanks for your collaboration!

PR #95, opened 2018-10-25
Added Google Dorking Capability 

NotoriousRebel:
Modified the tool to allow for the user to specify if they want to utilize google dorks. With a text file of 200 dorks will go through them and do get requests for each of them.  (depending on the limit to) Also, added a list of user agents to use when doing requests to decrease the chance of getting flagged by Google in googlesearch class.


laramies:
Great thanks, @NotoriousRebel !

PR #97, opened 2018-10-25
Update pgpsearch.py

blshkv:
I'm running a test with the standard example:
$ python2 theHarvester.py  -d microsoft.com -b pgp
However, the ```pgp.rediris.es``` server is dead.

This fix in this pull simply prints an error message to show the problem instead of returning 0 results. However, you might want to change the server to pgp.mit.edu as well (currently commented out).


laramies:
Thanks! I will change the server to use the MIT one.

Issue #96, opened 2018-10-25
shodan


Andrewblood:
When the server have more adresses it searches to fast with shodan
That's the output:

[-] Shodan DB search (passive):
 Request rate limit reached (1 request/ second). Please wait a second before trying again and slow down your API calls.

On https://developer.shodan.io/api stands also Note: All API methods are rate-limited to 1 request/ second. please fix that or help me to fix that.

jzold:
Hi Andrew,

There is a relatively quick fix for this, I'll take a look shortly

J./

laramies:
Closing this issue as @jzold provided a workaround.

PR #98, opened 2018-10-25
fix for issue #96: 2 seconds delay added to shodan search

jzold:
Added 2 seconds delay for shodan searches

https://github.com/laramies/theHarvester/issues/96


laramies:
Thanks for the workaround. I highly recommend getting a personal Key instead of using the one included in theHarvester :) will be easier to control the rate.

Issue #100, opened 2018-11-02
[New Feature - Censys]


jzold:
Hi all,

I've added a new search engine (Censys) to theHarvester to make it even more powerful.

For now the Censys module does the following:
- Gets all the IP addresses for a domain (limited to 100 results)
- Gets all the hostnames associated with that domain (limited to 100 results)

The censys module uses BeautifulSoup4 for parsing.

Notes: the limit is 10 searches/day without a censys account

Further improvement ideas:
1. Extend the Censys parser to include open ports, geographical locations, AS numbers, certificate CNs
2. Extend the Censys parser to alert when the limit is reached
3. Parameterize the limit search (currently it's hard coded into a while loop)
4. Implement authentication to Censys

Enjoy :)

J./

PR #101, opened 2018-11-02
[New Feature - Censys]

jzold:
Hi,

This is a new module (Censys) implementation pull request (ref: https://github.com/laramies/theHarvester/issues/100)

Any questions please let me know

Thanks
J./


NotoriousRebel:
Wouldn't you want to do a check in theHarvester.py to try and import BS4 and if they don't have it print the usage message and exit the program? Moreover, how difficult would it be to do the parsing without BS4?

jzold:
Hi Matt,

Good catch on the import, I only added it to requirements but checking the module from code also makes sense :) I added it already to a new pull request.

Using BS4 gives an opportunity to reduce complexity of theHarvester, reduces blast radius of code changes and increases modularity. There are a lot of new search engines using various and different formats that need parsing somehow. At the moment myparser does the parsing and factoring in the various formats search engines are using (XML/JSON/etc.), existing search engines changing their code and output formats...the complexity of myparser will increase. For example, if/when a new search engine is added to theHarvester it may require modifying myparser - which implicitly increases the risk of breaking those engines - which makes testing and code merges more challenging.

Using BS4 decouples parsing, reduces dependencies on the myparser library and potentially simplifies it. BS4 already supports HTML5 tags and various other formats (eg. XML, JSON, etc.). Using BS4 with a parser interface for a new engine also gives the additional benefit of reducing blast radius to that engine only. If the HTML code changes you don't have to tinker with myparser and risk breaking other engines in theHarvester.

Any other questions please let me know

Thanks
J./


Wouldn't you want to do a check in theHarvester.py to try and import BS4 and if they don't have it print the usage message and exit the program? Moreover, how difficult would it be to do the parsing without BS4?

jzold:
any other opinions about this one or any feedback?

NotoriousRebel:
Overall, it looks fine; however, make sure if the user wants to use every search engine that you keep the emails sorted and as a set to remove duplicates.

laramies:
Thanks @jzold and @NotoriousRebel  for this contribution, regarding BS4, makes sense to revise the parser to reduce complexity and improve detection. Regarding Cencysparser.py shouldn't this be a generic parser based on BS4, like myparser.py?  or even include it in myparser.py?

jzold:
@laramies  I thought about this myself :) If censysparser would be part of myparser it would follow the same pattern as it did before. My thoughts at that point were:

1. Adding censysparser to myparser: after a while (and potentially when implementing other search engines using BS4 ) myparser would become quite large and a bit  more challenging to maintain. This also means porting to python3 becomes potentially more challenging 
2. Censysparser is strongly dependent on the Censys website HTML and is prone to changes. I wanted to reduce and limit the impact and treating it as a separate module and not mess with myparser :) Eg. when theHarvester is run with "-all"  it would "bomb out" and interfere other search engine outputs if it was part of myparser. Separating it would not impact or interfere with theHarvester search results and it's easier to handle and manage Exceptions and debugging as well. 
3. Implementing a generic parser (like myparser) that uses BS4 would result in the same challenges as point1&point2 if other search engines are implemented into theHarvester using BS4

Any thoughts or suggestions? Maybe I'm over complicating this?

J./

laramies:
Let's try it that way, and we can review it when porting to Python3

PR #102, opened 2018-11-05
Added Hunter Search Engine

NotoriousRebel:
I added the hunter search engine and made it so that the user just needs to run it with their api key. I also modified the google search class that if they are using google dorks to increase the speed of it.


NotoriousRebel:
Reworking engine to flow better, temporarily closing!

NotoriousRebel:
I modified the integration of the search engine to be more logical and to not confuse the user.

laramies:
Great, thanks @NotoriousRebel

PR #106, opened 2018-11-26
fixed local variable error being referenced before defined in googleCSE search

NotoriousRebel:
Also removed a whitespace.


NotoriousRebel:
got fixed, closing!

PR #112, opened 2018-12-06
Censys page search fix

jzold:
This is a fix for the censys plugin. The fix minimizes the number of HTTP GETs to search result pages and eliminates the built in search limit.


jzold:
FIX: all censys search results (IP addresses and hostnames) are saved to SQLite

PR #118, opened 2018-12-16
Fixed program crashing if using search engine that does not define all_ip

NotoriousRebel:
Also made print statements more logical!


NotoriousRebel:
Closing since patched!

PR #123, opened 2018-12-18
Implemented uniform randomization and delay across plugins and fixed minor bugs.

NotoriousRebel:
Also made spacing more aesthetically appealing.


NotoriousRebel:
If you look at the second to last commit I accidentally pushed my private API key for hunter, it has been deleted and I generated a new key.

PR #124, opened 2018-12-19
Reporting model v0.4

jzold:
Changes:
1. Added latest scan with plugins used and results for each plugin
2. Added previous scan results for same domain theHarvester was run with plugins
3. Changed the bar chart for the latest scan results
4. Fixed formatting issues in the generated HTML
5. BUGFIX: return meaningful data when theHarvester runs the first time/day for the previous scan results


leebaird:
Why was this committed to the repo before my PR, which is older and fixes a lot of syntax issues?

laramies:
My bad @leebaird I was working on the Reporting with @jzold  and I merged that one thinking was the latest. I will merge yours and solve the conflicts.

PR #132, opened 2018-12-23
censys plugin fix

jzold:


leebaird:
In censysparser.py, add a space in front of the + on line 48.

PR #140, opened 2018-12-27
Code review of lib directory.

leebaird:


L1ghtn1ng:
Just a tip for you Lee that you do not need to make the files all 755 if they are lib files that have functions that just run by being imported into another main script like theharverster.py for example __init__.py files do not need to be 755 either as they just tell the python interpreter to treat that directory as module

leebaird:
Thanks man.

PR #136, opened 2018-12-27
 [Security] Bump requests from 2.18.4 to 2.21.0 

L1ghtn1ng:
Bumps [requests](https://github.com/requests/requests) from 2.18.4 to 2.21.0. **This update includes security fixes.**
- [Release notes](https://github.com/requests/requests/releases)
- [Changelog](https://github.com/requests/requests/blob/master/HISTORY.md)
- [Commits](requests/requests@v2.18.4...v2.21.0)

PR #141, opened 2018-12-28
Misc changes

L1ghtn1ng:
I have updated a manor of things to make it pep8 compliant and tidy up the root tree by putting the parsers in there on directory, sorted out line endings to be CRLF. I need **#136 to be merged first** before this PR can be merged


NotoriousRebel:
Two things:

1. In regards to pip installing things, what if the person is still using python2.7 can they still do pip3?
2. In line 141 when it creates the set of supported engines what's with the two empty lines?

Other than that it looks great!

L1ghtn1ng:
1. No, that is to install the dependencies for py3 only, pip is for py2

2. That will be down to pep8

Get Outlook for Android<https://aka.ms/ghei36>

PR #150, opened 2019-01-05
Added DuckDuckGo and updated filter.

NotoriousRebel:
Title says it all, for duckduckgo I utilized their API and parsed the json then crawled the remaining urls.


NotoriousRebel:
Also updated dockerfile for python3.6

PR #151, opened 2019-01-06
Misc changes

L1ghtn1ng:
Have done some refactoring, pep8 fixes all over the place, add some missing engines for the all flag and format fixes


L1ghtn1ng:
Also we need to stick to using single quotes through the code base and only have double when needed

PR #173, opened 2019-01-28
Removed additional google-profiles code.

leebaird:


leebaird:
Did not mean to delete api-keys.yaml

L1ghtn1ng:
Do a git revert and the hash and you will be able to get it back then you will have to do a few changes again

PR #176, opened 2019-02-04
Normalize whitespace and add gitattributes. Fixes #175.

### LINKED ISSUE #175: Mixed line endings
The source code currently contains a mixture of Unix (LF) and Windows (CRLF) line endings, which breaks the codebase. It appears at least some of the problems stem from d7d7d6a61e6c562a8f4924e4c4c32c16870218b5 when a bunch of CRLF endings was improperly committed to the repo.

Literal CRLF endings should never be committed to git for text files. git always works with LF internally, optionally converting to any other line endings in the working directory on checkout according to the `core.eol`, `core.autocrlf` and `core.safecrlf` settings.

The default setting of `core.eol` is `native` which is enough to ensure that git converts all line endings in text files to those appropriate for a given platform on checkout. A `.gitattributes` file should also be added to the repo to let git know which files are text and which are binary. This ensures that everyone gets the appropriate result regardless of their settings.

dkasak:
This fixes #175.


NotoriousRebel:
Thanks for the PR this is being looked at!

PR #187, opened 2019-02-14
Removed PGP source.

leebaird:


jinensetpal:
Why was support for PGP removed?

PR #199, opened 2019-02-21
Multiple data sources support

YuriyPobezhymov:
Hello.
Adds support for multiple data sources with simple syntax:
`theharvester -d microsoft.com -b 'pgp, virustotal'`
theharvester was confused to me when I tried to pass '-b' arguments with space and without quotes.

### DIFF
--- a/theHarvester.py
+++ b/theHarvester.py
@@ -81,7 +81,7 @@ def start():
     word = args.domain
 
     if args.source is not None:
-        engines = set(args.source.split(','))
+        engines = set(map(str.strip, args.source.split(',')))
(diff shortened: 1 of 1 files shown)


NotoriousRebel:
How it currently works is the user runs the tool as ```theharvester -d microsoft.com -b hunter,intelx``` it just splits on commas. would it be easier to remove spaces if they exist?

YuriyPobezhymov:
Possibly you are right, but stable version contains help that have spaces between data sources and that was not clear for me before I took a look into code.
That's my small propose.

PR #200, opened 2019-02-25
Use project requirements file to resolve dependencies.

leonjza:
The original `Dockerfile` installation was missing the `pyyaml` dependency in the `pip3 install` phase. Updating the container build to simply use the projects `requirements.txt` file both solves this issue as well as any future potential dependency changes.


NotoriousRebel:
Thanks for the PR!

PR #206, opened 2019-02-28
adds exclude flag and remove googleCSE from __init__

as77c:
Fixed a run time error by removing googleCSE from __init__.
Added a new flag -x to be used with -b all to exclude some sources.


leebaird:
I fixed this issue in the last commit: discovery/__init__.py

as77c:
> I fixed the run time error in the last commit.

Hi leebaird, this is strange since i could not get it run until i fixed __init__
screenshot

NotoriousRebel:
Thank you for the PR, this being looked at.

NotoriousRebel:
@as77c It should error out properly, are you using the most up to date version?

as77c:
@NotoriousRebel yes, it is last version

NotoriousRebel:
@as77c Why do you think it would be handy? If you do -b all why would you want to then exclude search engines? You can always just do -b searchengine1,searchegine2,searchengine3,...

as77c:
@NotoriousRebel As you know the search engines option is growing fast. I thought this will ease the problem from excluding some engines.

@L1ghtn1ng Thanks for your comments. PR is updated

theHarvester.py
```diff
@@ -30,7 +30,17 @@
 
 Core.banner()
 
-
+def modified_soruce(engines, excluded_engines):
+    engines = Core.get_supportedengines()
+    engines.remove('all')
+    excluded_engines = set(map(str.strip, excluded_engines.split(',')))
+    for exclude in excluded_engines:
```
NotoriousRebel: Instead of doing this for loop you can simply do engines = engines.difference(excluded_engines). This will subtract the two sets and is much cleaner. 
Ex: engines= {'Google', 'duckduckgo', 'baidu'}
      excluded_engines = {'baidu'}
      print(engines.difference(excluded_engines))
      {'Google','duckduckgo'}

as77c: Updated thanks!

as77c:
@L1ghtn1ng please check the updated PR.

L1ghtn1ng:
Will have a look at this tonight

theHarvester.py
```diff
@@ -94,7 +89,7 @@ def start():
     if args.source is not None:
         engines = set(map(str.strip, args.source.split(',')))
         if args.source == 'all' and args.exclude is not None:
-            engines = modified_soruce(engines, args.exclude)
+            engines = modified_soruce(args.exclude)
```
L1ghtn1ng: you have a typo, s/soruce/source/g

theHarvester.py
```diff
@@ -30,16 +30,11 @@
 
 Core.banner()
 
-def modified_soruce(engines, excluded_engines):
+def modified_soruce(excluded_engines):
```
L1ghtn1ng: see below comment

PR #228, opened 2019-03-23
Added data files to setup.py to handle files in wordlists directory

NotoriousRebel:
Title says it all :)


NotoriousRebel:
Also, modified netcraft and made a small change to dnsdumpster plugin

PR #229, opened 2019-03-24
Update setup.py

blshkv:
install theHarvester binary and register it as a module


setup.py
```diff
@@ -28,4 +28,9 @@
         ]
          )
     ]
+    entry_points={
+        'console_scripts': [
+            'theHarwester = theHarvester.theHarvester:start'
```
L1ghtn1ng: Typo

blshkv:
ok, I'm done for today. The current patch fixes the installation issue so that it can be installed as a system python module. However, all output still must be redirected to a homedir so that a regular user could run it

PR #237, opened 2019-04-16
setup.py: do not install tests

blshkv:
I'm getting the following error:
```Package installs "tests" package which is forbidden and likely a bug in the build system.```

This patch fixes it


NotoriousRebel:
Looks good :)

PR #239, opened 2019-04-19
Travis changes and removed win32dns.py

L1ghtn1ng:
Title says it all and I have also renaned this version to be 3.1.0.dev0 as I would say the changes when we are happy to do a release would allow for a major version bump.


L1ghtn1ng:
I still want to do a few more changes to travis before this gets merged

L1ghtn1ng:
This is ready to merge, it does make the build take longer but, should catch more issues before it gets merge into master

PR #270, opened 2019-07-21
Fixes for bugs and updated deps

L1ghtn1ng:
@NotoriousRebel If you could fork my fork and help with the crt.sh refactor and submit a PR to my upstream this then would update the PR when you submit it to my dev branch.

That would be great if you could do that ASAP. THanks


L1ghtn1ng:
Just need to do one last change to make sure that we now test crt.sh as part of the ci I need to do

L1ghtn1ng:
@NotoriousRebel I am going to merge this as least it fixes a lot of stuff but we needed to find a suitiable way of error handling the crtsh.py as when it times out it crashes out completely of the script which only happens on domains that are huge and have a massive amount of hosts

PR #304, opened 2019-09-19
fixed code flow to run google_workaround for googlesearch linkedinsearch

VKen:
The new code added in commit https://github.com/laramies/theHarvester/commit/f702f56d63bb70e531963bdbd3ec0fe5df2a8f56#diff-71d73163cccca7ec94a01732e74da4ceR34

Seems to not run the `google_workaround` function at all for `googlesearch.py` and `linkedinsearch.py`.

```python
self.results = r.text
if search(self.results):
    try:
        if isinstance(search(self.results), bool):
            print('Google is blocking your ip and the workaround, returning')
            return
        else:
            self.results = google_workaround(urly)
    except BaseException:
        pass
```
Not sure about the intent, but it seems weird to have a work around but not able to use it.

Looking at the `search` function in `constants.py`, it will always return a boolean object (as below). Hence, the above code will always print the blocked-message and return.

```python
def search(text: str) -> bool:
    # Helper function to check if Google has blocked traffic.
    for line in text.strip().splitlines():
        if 'This page appears when Google automatically detects requests coming from your computer network' in line \
                or 'http://www.google.com/sorry/index' in line or 'https://www.google.com/sorry/index' in line:
            # print('\tGoogle is blocking your IP due to too many automated requests, wait or change your IP')
            return True
    return False
```

The new (but incomplete) `google_workaround` returns boolean `True` on checking for google block content.

```python
def google_workaround(visit_url: str) -> Union[bool, str]:
...
    if search(returned_html):
        # indicates that google is serving workaround a captcha
        # TODO rework workaround with more websites to send requests on our behalf or utilize proxies option in request
        return True
```

Therefore, I'm making a guess the intent and flow of the search logic to be:

1. check result for blocking
2. try workaround
3. check workaround for blocking

The result is this:
```python
self.results = r.text
if search(self.results):
    try:
        self.results = google_workaround(urly)
        if isinstance(self.results, bool):
            print('Google is blocking your ip and the workaround, returning')
            return
    except Exception:
        # google blocked, no useful result
        return
```

### DIFF
--- a/theHarvester/discovery/googlesearch.py
+++ b/theHarvester/discovery/googlesearch.py
@@ -30,13 +30,13 @@ class SearchGoogle:
         self.results = r.text
         if search(self.results):
             try:
-                if isinstance(search(self.results), bool):
+                self.results = google_workaround(urly)
+                if isinstance(self.results, bool):
                     print('Google is blocking your ip and the workaround, returning')
                     return
-                else:
-                    self.results = google_workaround(urly)
             except Exception:
-                pass
+                # google blocked, no useful result
+                return
         time.sleep(getDelay())
         self.totalresults += self.results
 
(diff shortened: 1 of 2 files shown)


L1ghtn1ng:
Matt can you please review this PR please

NotoriousRebel:
This is being looked at :)

VKen:
Thank you

PR #323, opened 2019-10-05
Simplify `unique` method in myParser

### LINKED ISSUE #316: Rework of unique method inside myparser.py
Inside theHarvester/parsers, the method unique inside myparser.py could use reworking as it is convoluted and could be improved upon.

cueo:
Fix #316

### DIFF
--- a/theHarvester/parsers/myparser.py
+++ b/theHarvester/parsers/myparser.py
@@ -149,8 +149,4 @@ class Parser:
         return urls
 
     def unique(self) -> list:
-        self.new = []
-        for iteration in self.temp:
-            if iteration not in self.new:
-                self.new.append(iteration)
-        return self.new
+        return list(set(self.temp))


cueo:
@NotoriousRebel is the order important for the `unique()` method?

NotoriousRebel:
@cueo Your changes look good! You should cast to a set before you cast to a list as a set is that makes the values unique ;) Will merge soon.

L1ghtn1ng:
@cueo Can you rebase this please as this PR has merge conflicts

L1ghtn1ng:
Also please implement @NotoriousRebel changes, thanks

cueo:
@L1ghtn1ng I have rebased the PR.
Also, I don't think @NotoriousRebel requested any changes, he mentioned that in order to make values unique I should cast to a `set` first which is what I am doing. Is it okay to be merged now?
Thanks!

cueo:
> @cueo Your changes look good! You should cast to a set before you cast to a list as a set is what makes the values unique ;) Will merge soon.

@NotoriousRebel One thing that we should be wary about is order of elements in `self.temp` might change on casting it to a `set`.

L1ghtn1ng:
It needs to be set then list not the other way round. That is what @NotoriousRebel was getting at?

cueo:
@L1ghtn1ng that's what I have done. Not sure if I'm missing something here.

`self.temp` is first cast to a set -> `set(self.temp)`
since we need to return a list, we convert that set to list -> `list(set(self.temp))`

L1ghtn1ng:
Ah okay now that makes sense to me it was ```list(set(self.temp))``` that was throwing me off

Issue #317, opened 2019-10-05
Reduce code repetition


NotoriousRebel:
Inside __main__.py there is a lot of code repetition as well as in the discovery modules for example, most modules have the same do_search method with a different list of urls and most have the same get_emails and get_hostnames functions. Inside __main__.py  for each engineitem it does very similiar function calls.  Most of these functions could be put inside constants.py or core.py and this would reduce a lot of code repetition.

L1ghtn1ng:
Core.py would be better
________________________________

maxerenberg:
I submitted a pull request for this issue. Sorry for all the mini commits - flake8 was giving me errors because it couldn't recognize transitive imports. This is my first time contributing so please let me know if I'm missing anything

maxerenberg:
I also removed 'vhost' from the -b options because it wasn't even being checked

maxerenberg:
@L1ghtn1ng Thank you for the feedback. Would it perhaps be a good idea to have one abstract base class with common methods for the Search engines, and have each engine from that? It might not necessarily reduce the total amount of code, but it would avoid having a gigantic if statement in __main__.py.

NotoriousRebel:
@maxerenberg At the moment there are two big cases of code repetition that should take highest priority. The first case is inside __main__.py when we find the engine we want to use, you notice it is the same formula in which it creates the discovery class gets the results then stores it in the database. We could for example move that to a new file or create new functions inside core.py. Furthermore, another big case is inside the discovery modules as in most cases we simply just generate the base urls make our requests and just concatenate the results to self.total_results. However, in almost every discovery module it makes a call to get _hostnames, get_emails, or get_allhostnames. That can be put into separate function/functions inside core.py.

There are multiple different ways to solve this issue; however, having core functions that are utilized across the tool should be put into core.py or maybe a new class for example.

maxerenberg:
@NotoriousRebel I noticed that most of the discovery classes have very similar methods which accept very similar sets of parameters. I think it would be a good idea to have one abstract base class from which all the discovery classes inherit. This way, we could group together common blocks of code and have one common interface for every discovery class. For example, we could have something like
```
class AbsractSearchClass:
    def do_search(self):
        """To be implemented"""
        pass

    def get_emails(self):
        rawres = myparser.Parser(self.total_results, self.word)
        return rawres.emails()

    def get_hostnames(self):
        rawres = myparser.Parser(self.total_results, self.word)
        return rawres.hostnames()

...

[...]

PR #330, opened 2019-10-06
grouped together common code blocks to reduce repitition

maxerenberg:
Trying to reduce code repetition as per https://github.com/laramies/theHarvester/issues/317


L1ghtn1ng:
@maxerenberg Thanks for the PR, unfortunately this does not meet the high standards that we require and is not done in a neat way. You have modified things that did not need touching like the API keys for example. We like to have things simple as much as possible to make things easy to maintain and keep the code looking tidy. This does not do that unfortunately. You also need to make sure you test things with flake8, mypy and pytest as well before submitting a PR.

You can resubmit a PR when done but for now this is getting closed.Also please take a look at the following https://github.com/laramies/theHarvester/blob/master/CONTRIBUTING.md

If you need more information please ask @NotoriousRebel or myself in the issue thread and we will try to explain it as clearly as possible. Before submitting a PR if you link to it in the issue thread we can have a look before hand. Good first try though :-)

Issue #334, opened 2019-10-12
Fix mypy errors for github module


L1ghtn1ng:
```
theHarvester/discovery/githubcode.py:59: error: Incompatible types in assignment (expression has type "List[Any]", variable has type "str")  [assignment]
theHarvester/discovery/githubcode.py:61: error: Incompatible return value type (got "Union[str, int]", expected "Optional[int]")  [return-value]
```

L1ghtn1ng:
With the new version of mypy we now have errors again 

```
theHarvester/discovery/githubcode.py:58: error: List item 0 has incompatible
type "None"; expected "str"  [list-item]
                pages = params.get('page', [None])
```

L1ghtn1ng:
I have fixed 1 error by changing the function to be an Optional[Any] but would love some help with this error

PR #337, opened 2019-10-13
fix mypy errors in discovery/githubcode.py

BlueBlazin:
Fix https://github.com/laramies/theHarvester/issues/334


L1ghtn1ng:
This is a duplicate of #338 I preferred the way it was done in that PR. Thanks for the contribution though and keep them coming :-)

PR #338, opened 2019-10-13
Fix mypy error for github module

### LINKED ISSUE #334: Fix mypy errors for github module

cueo:
Fix #334


L1ghtn1ng:
THanks for this

PR #340, opened 2019-10-14
Externalise code repetition to method

### LINKED ISSUE #317: Reduce code repetition

cueo:
Fixes #317 

This PR only touches the logic inside search engines, externalising the logic to store hosts, emails, ips to a nested method.

### DIFF
--- a/theHarvester/__main__.py
+++ b/theHarvester/__main__.py
@@ -64,7 +64,7 @@ def start():
     shodan = args.shodan
     start = args.start  # type: int
     takeover_check = False
-    trello_urls = []
+    all_urls = []
     vhost = []
     virtual = args.virtual_host
     word = args.domain  # type: str
@@ -110,7 +110,7 @@ def start():
             email_list, host_names, urls = search_engine.get_results()
             all_emails.extend(email_list)
             host_names = filter(host_names)
-            trello_urls.extend(filter(urls))
+            all_urls.extend(filter(urls))
             all_hosts.extend(host_names)
             db.store_all(word, all_hosts, 'host', source)
             db.store_all(word, all_emails, 'email', source)
@@ -408,7 +408,7 @@ def start():
             print(host)
         host_ip = [netaddr_ip.format() for netaddr_ip in sorted([netaddr.IPAddress(ip) for ip in ips])]
         db.store_all(word, host_ip, 'ip', 'DNS-resolver')
-    length_urls = len(trello_urls)
+    length_urls = len(all_urls)
     if length_urls == 0:
         if len(engines) >= 1 and 'trello' in engines:
             print('\n[*] No Trello URLs found.')
@@ -416,7 +416,7 @@ def start():
         total = length_urls
         print('\n[*] Trello URLs found: ' + str(total))
         print('--------------------')
-        for url in sorted(trello_urls):
+        for url in sorted(all_urls):
             print(url)
 
     # DNS brute force
--- a/theHarvester/__main__.py
+++ b/theHarvester/__main__.py
@@ -126,9 +126,9 @@ def start():
                     print(usr)
         if store_links:
             links = search_engine.get_links()
-            db.store_all(word, links, 'name', 'linkedin')
+            db.store_all(word, links, 'name', engineitem)
             if len(links) == 0:
-                print('\n[*] No links found Linkedin.\n\n')
+                print('\n[*] No links found.\n\n')
             else:
                 print(f'\n[*] Links found: {len(links)}')
                 print('---------------------')
@@ -176,12 +176,7 @@ def start():
                     from theHarvester.discovery import certspottersearch
                     try:
                         certspotter_search = certspottersearch.SearchCertspoter(word)
-                        certspotter_search.process()
-                        hosts = filter(certspotter_search.get_hostnames())
-                        all_hosts.extend(list(hosts))
-                        all_hosts.extend(hosts)
-                        db = stash.stash_manager()
-                        db.store_all(word, all_hosts, 'host', 'certspotter')
+                        store(certspotter_search, engineitem, None, store_host=True)
                     except Exception as e:
                         print(e)
 
--- a/theHarvester/__main__.py
+++ b/theHarvester/__main__.py
@@ -101,7 +101,7 @@ def start():
         if store_ip:
             ips_list = search_engine.get_ips()
             all_ip.extend(ips_list)
-            db_stash.store_all(word, ips, 'ip', source)
+            db_stash.store_all(word, all_ip, 'ip', source)
         if store_data:
             data = filter(search_engine.get_data())
             all_hosts.extend(data)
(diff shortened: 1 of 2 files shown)


cueo:
https://github.com/laramies/theHarvester/blob/master/theHarvester/__main__.py#L163-L174

In this code snippet, we retrieve hosts and emails but persist hosts for both host and email. Is this on purpose? If so, this PR would need to be revisited.

NotoriousRebel:
@cueo Thank you for the PR, it is currently being reviewed. In regards to your question above there seems to be a mistake in which depending on the engine we do db.store_all(word, all_hosts, 'email', 'module`) when we should be doing  db.store_all(word, all_emails, 'email', 'module') . Making a reference to all_emails as well as all_hosts.

theHarvester/__main__.py
```diff
@@ -69,6 +69,56 @@ def start():
...
+            db_stash.store_all(word, ips, 'ip', source)
+        if store_data:
+            data = filter(crtsh_search.get_data())
+            all_hosts.extend(data)
+            db.store_all(word, all_hosts, 'host', source)
+        if store_results:
+            emails, hosts, urls = search_engine.get_results()
+            all_emails.extend(emails)
+            hosts = filter(hosts)
+            global trello_urls
```
NotoriousRebel: Is the use of the global keyword truly necessary?

cueo:
> It seems that store_host and store_emails are always true. Should the default argument be false? Is it necessary to check if store_host, is always true?

@NotoriousRebel There are a couple of instances where hosts and emails are not stored, eg: linkedin and twitter.

theHarvester/__main__.py
cueo: we need to declare the variable as `global` because we are assigning it and then accessing it [here](https://github.com/laramies/theHarvester/blob/master/theHarvester/__main__.py#L499)

if we change L97 to `trello_urls.extend(filter(urls)` then we can skip the `global` keyword

NotoriousRebel: I would prefer not using global unless absolute necessary, in this case doing will suffice.

cueo:
As mentioned in #317, before we move the common code to a different class, I think this is right way to start because there are some variables which are referenced later in the code. We might need to do some clean up before completely moving out the common code.

NotoriousRebel:
@cueo In regards to my earlier comment after closer analysis the way you are doing it is fine as at the moment take for example the linkedin module does not store_hosts. Furthermore, this is a great start in reducing code repetition, once @L1ghtn1ng and I do more testing this should be ready to merge. Would it be possible to add any type hinting in the meantime?

cueo:
> @cueo In regards to my earlier comment after closer analysis the way you are doing it is fine as at the moment take for example the linkedin module does not store_hosts. Furthermore, this is a great start in reducing code repetition, once @L1ghtn1ng and I do more testing this should be ready to merge. Would it be possible to add any type hinting in the meantime?

@NotoriousRebel Sure, I'll get started on type hinting for the newly added method.

theHarvester/__main__.py
```diff
@@ -69,6 +69,56 @@ def start():
...
+        if store_links:
+            links = linkedin_links_search.get_links()
+            db.store_all(word, links, 'name', 'linkedin')
+            if len(links) == 0:
+                print('\n[*] No links found Linkedin.\n\n')
+            else:
+                print(f'\n[*] Links found: {len(links)}')
+                print('---------------------')
+                for link in sorted(list(set(links))):
+                    print(link)
```
L1ghtn1ng: I am not to sure about having this in this function. I do have another module that I will be creating at some point soon that will be return links and that will be its soul purpose but feel like there is a better way to make it agnostic. @NotoriousRebel Thoughts on this?

cueo: Totally agree that this should be agnostic. However, I didn't want to touch too much code around for refactoring and wanted to solely reduce those code repetitions. I can work on externalizing the method in next PR.

Will definitely add documentation and see if any unit tests can be added.

cueo:
@L1ghtn1ng is it okay if I can add unit tests in a separate PR?
I have added DocString for the method. Let me know if you want me to add more details.

L1ghtn1ng:
Would like unit tests to be with this this PR as I want new code like this to have it be part of the same pr

cueo:
@L1ghtn1ng there is no unit test for existing code.
I'd need to write tests for all existing logic while mocking the dependencies. I will try to take it up over weekend. Didn't want this PR to be blocked because of that.
If you still think it's better to have it as part of this PR, will try to do that.

L1ghtn1ng:
Would be better to have it as part of this PR want to avoid breaking this if possible

theHarvester/__main__.py
```diff
@@ -125,7 +125,7 @@ def store(search_engine: Any, source: str, process_param: Any = None, store_host
                 for usr in sorted(list(set(people_list))):
                     print(usr)
         if store_links:
-            links = linkedin_links_search.get_links()
+            links = search_engine.get_links()
             db.store_all(word, links, 'name', 'linkedin')
```
L1ghtn1ng: Can you get it to not be linkedin as previously mentioned other modules that get links would reference linkedin if this is used so can you change it to the engine name that is requesting to save the links

theHarvester/__main__.py
```diff
@@ -125,7 +125,7 @@ def store(search_engine: Any, source: str, process_param: Any = None, store_host
...
             if len(links) == 0:
                 print('\n[*] No links found Linkedin.\n\n')
```
L1ghtn1ng: Replace linkedin with the engine name that is searching for links

cueo: Thanks for spotting that!

cueo:
@L1ghtn1ng Added skeleton for unit tests.
Now just need to copy paste the same code for other search engines as well.

cueo:
@L1ghtn1ng unit tests now mostly covers all the use cases.
Please review

cueo:
@NotoriousRebel is it at go merge now?

L1ghtn1ng:
I have spoken to @NotoriousRebel and we would like the store() to be moved into core.py and become a class method

cueo:
@L1ghtn1ng @NotoriousRebel

As mentioned earlier, before we move the common code to a different class, I think this is right way to start because there are some variables which are referenced later in the code. We might need to do some clean up before completely moving out the common code.

L1ghtn1ng:
If that is what needs to be done then are you able to do it as part of this PR? I would be fine with that. Also for the unittest any reason for using unittest instead of pytest?

cueo:
I'd need to refractor some other parts of the code which I need to
look into. I would really appreciate if I could pick that up as part of the
next PR. As part of that, we could also look into writing a common abstract
class which is extended by different search engine classes.

I am using pytest for writing test and unittest for mocking dependencies. I
liked the ease of using `@patch` for mocking compared to `monkeypatch`.

L1ghtn1ng:
Okay fair enough and I am happy with that and so is @NotoriousRebel, massive thanks for this and good job :-)

PR #342, opened 2019-10-15
Fix CI error

cueo:
Remove non-existing directory reference


cueo:
@L1ghtn1ng please review

Issue #363, opened 2019-11-11
Odd debug showing before the banner


digininja:
**Feature Request or Bug or Other**
Other

**Describe the feature request or bug or other**
Each time I run the app I get the message "table results already exists" before the banner. It looks like debug that has been left in somewhere but I can't find a reference to that text anywhere in the app itself so it may be from one of the dependencies.

**To Reproduce**

Simply run the app:

```
$ pipenv run python3.7 ./theHarvester.py 
table results already exists

*******************************************************************
*  _   _                                            _             *
* | |_| |__   ___    /\  /\__ _ _ ____   _____  ___| |_ ___ _ __  *
* | __|  _ \ / _ \  / /_/ / _` | '__\ \ / / _ \/ __| __/ _ \ '__| *
* | |_| | | |  __/ / __  / (_| | |   \ V /  __/\__ \ ||  __/ |    *
*  \__|_| |_|\___| \/ /_/ \__,_|_|    \_/ \___||___/\__\___|_|    *
*                                                                 *
* theHarvester 3.1.1dev3                                          *
* Coded by Christian Martorella                                   *
* Edge-Security Research                                          *
* cmartorella@edge-security.com                                   *
*                                                                 *
******************************************************************* 


usage: theHarvester.py [-h] -d DOMAIN [-l LIMIT] [-S START] [-g] [-p] [-s]
                       [-v] [-e DNS_SERVER] [-t DNS_TLD] [-n] [-c]
                       [-f FILENAME] [-b SOURCE]
theHarvester.py: error: the following arguments are required: -d/--domain
```

**Expected behaviour**
No debug before the banner

**System Information (System that tool is running on):**
Ubuntu 18.04 up-to-date with the latest git code and a pip environment just created.

**Additional context**
It might not be your problem, but if not, I've no idea where I should be looking.

NotoriousRebel:
This may be with when the program first interacts with SQLite will look into it.

digininja:
I have updated from an older version so it may be that the old database
structure needs a tweak to bring it up to date.

L1ghtn1ng:
@digininja We have noticed this ourselfs but not sure what it is, you idea sounds very plausible but as @NotoriousRebel  says we will have a look into that, but for now it does not causing any issues from what we have observed

digininja:
Didn't seem to cause a problem beyond looking odd.

Happy to help debug if needed.

L1ghtn1ng:
[...]

PR #366, opened 2019-11-13
only create table if it doesn't exist

digininja:
Fixing issue https://github.com/laramies/theHarvester/issues/363


NotoriousRebel:
@digininja thanks for the PR, merging now :)

PR #385, opened 2020-01-22
remove url scheme in genericClean Parser method

VoltBoy:
Hostnames output contains scheme without URL specific symbols like "://", so when the script is finished it's execution we get subdomains in wrong format, example: "httpsubdomain.test.com" or "httpssubdomain.test.com".

I fixed this error by passing possible url schemes in genericClean Parser method.

### DIFF
--- a/theHarvester/parsers/myparser.py
+++ b/theHarvester/parsers/myparser.py
@@ -13,7 +13,7 @@ class Parser:
             .replace('%2f', '').replace('%3a', '').replace('<strong>', '').replace('</strong>', '')\
             .replace('<wbr>', '').replace('</wbr>', '')
 
-        for search in ('<', '>', 'https', 'http', ':', '=', ';', '&', '%3A', '%3D', '%3C', '/', '\\'):
+        for search in ('<', '>', 'https://', 'http://', ':', '=', ';', '&', '%3A', '%3D', '%3C', '/', '\\'):
             self.results = self.results.replace(search, ' ')
 
     def urlClean(self):
--- a/theHarvester/parsers/myparser.py
+++ b/theHarvester/parsers/myparser.py
@@ -13,7 +13,7 @@ class Parser:
             .replace('%2f', '').replace('%3a', '').replace('<strong>', '').replace('</strong>', '')\
             .replace('<wbr>', '').replace('</wbr>', '')
 
-        for search in ('<', '>', ':', '=', ';', '&', '%3A', '%3D', '%3C', '/', '\\'):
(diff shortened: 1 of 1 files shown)


L1ghtn1ng:
Can you please give us reproduce steps for a domain that you get this error with?

VoltBoy:
> Can you please give us reproduce steps for a domain that you get this error with?

python theHarvester.py -d pochtabank.ru -b bing
Here's one example of when this error is occured. It happens not only with bing, but with google, yahoo and other search engines too.

NotoriousRebel:
@VoltBoy Thank you for making this PR, it might make more sense if we only remove http/https if we check if the string starts with http/https. What if the url contains http/https?

VoltBoy:
> @VoltBoy Thank you for making this PR, it might make more sense if we only remove http/https if we check if the string starts with http/https. What if the url contains http/https?

Yeah, there will be an error if url contains http or https at the beginning. So I just added a '://' symbols to http and https tuple elements.

L1ghtn1ng:
@NotoriousRebel Think we can merge this?

VoltBoy:
I found a much more interesting moment about this issue. 
genericClean Parser method contains a multiple replacements at the beginning, it also replaces a slash symbol to nothing.

`.replace('%2f', '')`

Because of it, users get a not valid subdomains in script output (if one of results from search engine contains links like 'somesite.com/*target domain*', after calling the genericClean method, self.results property of Parser contains 'somesite.com' concatenated with target domain). 

I suggest to replace slash symbol to space, like it already has been done in for cycle after multiple replacements, there it is:
`for search in ('<', '>', 'https://', 'http://', ':', '=', ';', '&', '%3A', '%3D', '%3C', '/', '\\'):
            self.results = self.results.replace(search, ' ')`

So, i've removed a `.replace('%2f', '')` statement at the beginning.

VoltBoy:
Now it's not necessary to replace 'http://' and 'https://', remove it from tuple in genericClean.

NotoriousRebel:
@VoltBoy how have you tested this? Looks good will merge soon.

VoltBoy:
@NotoriousRebel  I tested changes by simple running the script. Since i've made changes, output become clearer with searchers that using myparser. Also i added a '%2f' replacement to space, just in case (if link from search engine will be url encoded).
There is a few other problems with results from intelx, suip, dogpile, netcraft. But it's another issues. I'am ready to make a contribution with bugfixes, if there is enough free time in the future.

PR #400, opened 2020-03-13
fix #397

### LINKED ISSUE #397: [BUG] XML/HTML output functionality can't handle paths with periods or relative paths
**Feature Request or Bug or Other**
Bug

**Describe the feature request or bug or other**
Hey folks!
I'm doing my OSCP labs, and I've been saving all of the work for each lab exercise under a folder for that exercise. So if the exercise is chapter 6.12.1.1 , exercise 1, I would want to run 

```
theHarvester -d megacorpone.com -b google -f 6.12.1.1/1.html
```
This works fine for generating the html file, but it places the xml file in `6.xml` in the directory where the command is run, not in the `6.12.1.1` directory I suspect that this is because of [this line](https://github.com/laramies/theHarvester/blob/3c4d66d1313d1a00d4e7bc01a4f3d7a1705618fc/theHarvester/__main__.py#L613)

```python
filename = filename.split('.')[0] + '.xml'
```
 I believe it would be better to write
```python
filename = '.'.join(filename.split('.')[0:-1]) + '.xml'
```
Here is a demo of it working in the python shell
```python
>>> filename="a.b/c.html"
>>> filename.split('.')[0]
'a'
>>> '.'.join(filename.split('.')[0:-1])
'a.b/c'
```

**Happy to submit a PR, just let me know if you approve of my solution**

**To Reproduce**
Steps to reproduce the behaviour:
1. Run `theHarvester -d megacorpone.com -b google -f 6.12.1.1/1.html`

**Expected behaviour**
1. an html file is produced in `6.12.1.1/1.html`
2. an xml file is produced in `6.12.1.1/1.xml`

**Actual behaviour**
1. an html file is produced in `6.12.1.1/1.html`
2. an xml file is produced in `6.xml`

**System Information (System that tool is running on):**
 - OS: Kali
 - Version: 3.1.0

derwentx:


derwentx:
Hey @NotoriousRebel  looks like one of the github actions is broken. 
```
Pull access denied for kalilinux/kali-linux-docker, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

L1ghtn1ng:
I have that fixed in my branch but still doing changes to it so the docker build works now will merge it over when finished

L1ghtn1ng:
Thanks for this

derwentx:
> Thanks for this

Thank YOU for keeping this going! :D

PR #404, opened 2020-03-20
Reverse dns lookup

moodule:
Hey, confinement has its virtues I finally got around to doing this :grin:

Hopefully you made the tool async, and pushed me in this direction!
It got a nice bump in speed compared to my previous draft :zap: 

# Processing Example

Suppose  we got ['35.181.144.171', '195.154.2.7', '35.181.144.38'] from the search engines.

The module will reverse each ip in:
- 35.181.144.0/24 => 35.181.144.0 à 35.181.144.255
- 195.154.2.0/24 => 195.154.2.0 à 195.154.2.255

And keep only the host names that contain the target "word" (like "github.com")

# Caveats

- even with the async boost, it's still thousands of queries: it takes several minutes
- the results are only added to the report, not the db stash
- the result matching is naive: looking for subdomains of "aaaa.com", I match "ccc.bbbaaaa.com"
- some subdomains are not registered / grouped on a single ip

# It's a WIP

Well I'm not even sure it does what you intended!
(I'm new in this field...)


L1ghtn1ng:
You can remove the from future code as the project is python3 only 3.7+ supported and can you fix the flake8 issues a the lgtm issues as well please

theHarvester/discovery/dnssearch.py
```diff
@@ -40,3 +59,191 @@ def process(self):
...
+        The last 2 digits of the ip can be set to anything,
+        they will be ignored.
+    callback: Callable.
+        Arbitrary postprocessing function.
+
+    Returns
+    -------
+    out: None.
+    """
+    __resolver = DNSResolver(timeout=4)
```
NotoriousRebel: While setting up the DNSResolver the user may also specifiy the dns server to make the requests for example with dns.resolver you can do [this](https://github.com/L1ghtn1ng/theHarvester/blob/dev/theHarvester/discovery/dnssearch.py#L13). With aiodns you can pass in a list of nameservers when instantiate the DNSResolver see [here](https://github.com/saghul/aiodns/blob/master/aiodns/__init__.py#L46) for more information. Please add that functionality to allow the user to pass a dns server into this function to make the requests.

theHarvester/discovery/dnssearch.py
```diff
@@ -40,3 +59,191 @@ def process(self):
...
+        A serialized ip range, like '1.2.3.0/24'.
+        The last digit can be set to anything, it will be ignored.
+
+    Returns
+    -------
+    out: list.
+        The list of IPs in the range.
+    """
+    try:
+        __network = IPv4Network(iprange, strict=False)
```
NotoriousRebel: Any reason why strict is set to False?

theHarvester/discovery/dnssearch.py
```diff
@@ -1,8 +1,27 @@
...
+Explore the space around known hosts & ips for extra catches.
+"""
+
+import dns
+import re
 import sys
-import dns.resolver
+
+from aiodns import DNSResolver
+from ipaddress import IPv4Network
```
NotoriousRebel: Maybe not now but perhaps down the road there are instances where the tool grabs IPv6 addresses will there be any support for them?

theHarvester/discovery/dnssearch.py
moodule: ok, should I add a CLI argument for the user?

theHarvester/discovery/dnssearch.py
moodule: yep: the search engines collects ips like '178.128.43.171' but the 4th number must be set to 0 when creating an IPv4Network object.

so instead of manually replacing the "171" by "0", I ask IPv4Network to be lenient and still create the network "178.128.43.0/24" itself

theHarvester/discovery/dnssearch.py
moodule: well technically it should be no problem, but I start from the ips returned by the search engines, still ipv4 so I didn't feel the need

moodule: Suppose we grab "2001:db8:85a3:8d3:1319:8a2e:370:7348", what range would you reverse?

moodule:
Why apple.com??
It's 50+k of console log to scroll + the testing time :sleeping: 

Just putting flake8 / mypy tests before the actual call would be helpful for noobies / restless people!

moodule:
I don't know about the docker image, it looks unrelated?

theHarvester/discovery/dnssearch.py
NotoriousRebel: This is already a CLI argument see [here](https://github.com/laramies/theHarvester/blob/master/theHarvester/__main__.py#L31)

moodule: oh wow I totally looked at this and missed it!..
ok this should do it.

L1ghtn1ng:
You can close each section in GitHub actions so you can skip sections and or hide certain output. Yeah do not worry about docker as I am sorting that

theHarvester/discovery/dnssearch.py
L1ghtn1ng: /52 maybe?

PR #425, opened 2020-05-07
[Fix Typo] Removed 2nd `t` in `statges` of README.md

matsawyer:
`stages` seems to be the intended wording


leebaird:
Thank you!

PR #430, opened 2020-05-12
Added rapiddns as new module

NotoriousRebel:
Also changed default timeout to 60 seconds.


leebaird:
Super fast getting this new plugin coded! Thanks

PR #550, opened 2020-10-05
Add /usr/local/etc as possible location for config files

alexmerkel:
The change make theHarvester search for `proxies.yaml` etc. in `/usr/local/etc/theHarvester/` directory in addition to `/etc/theHarvester` and the current directory. This allows system-wide configs to be installed without root privileges, e.g. via Homebrew. 

Additionally, I refactored the code in `core.py` at bit. Now there is one method to read the `api-keys.yaml` file and all service-specific methods just call this function and pass on the right key from the returned object.

### DIFF
--- a/theHarvester/discovery/dnssearch.py
+++ b/theHarvester/discovery/dnssearch.py
@@ -31,11 +31,15 @@ class DnsForce:
         # self.dnsserver = [dnsserver] if isinstance(dnsserver, str) else dnsserver
         self.dnsserver = list(map(str, dnsserver.split(','))) if isinstance(dnsserver, str) else dnsserver
         try:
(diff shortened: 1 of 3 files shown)


L1ghtn1ng:
@alexmerkel Thanks for this and this is call that you have added homebrew support, the other thing that you will need to look at though is the stash.py for the sqlite DB or does the current way we do it work? it would need to stay that way but like how you have done the api-keys if it needs to be modified

alexmerkel:
The location of the sqlite database is fine imo. It is written to a user controlled location (`~/.local`), so installing (and running) theHarvester as a non-root user works. I added the other file locations because the original ones are in a location (`/etc`) typically not writable for normal users.

PR #623, opened 2020-12-14
Improve error output handling and add censys, omnisint modules

L1ghtn1ng:


api-keys.yaml
```diff
@@ -1,15 +1,19 @@
...
 
   github:
     key: 
 
   hunter:
-    key: 
+    key:
 
   intelx:
-    key: 9df61df0-84f7-4dc7-b34c-8ccfb8646ace
```
thehappydinoa, line 12: You might want to BFG this

PR #674, opened 2021-02-12
Dockerfile multiple fixes inspired by hadolint

aminvakil:
https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache
> This avoids the need to use --update-cache and remove /var/cache/apk/* when done installing packages.

https://github.com/hadolint/hadolint/wiki/DL3042#rationale
> Once a package is installed, it does not be installed and the Docker cache can be leveraged instead. Since the pip cache makes the images larger and is not needed, it's better to disable it.

> It's better to use `./*glob*` or -- `*glob*` so names with dashes won't become options.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM python:alpine3.11
 RUN mkdir /app
 WORKDIR /app
 COPY . /app
-RUN apk add --update build-base libffi-dev libxml2-dev libxslt-dev
-RUN pip3 install -r requirements.txt
-RUN chmod +x *.py
+RUN apk add --no-cache build-base libffi-dev libxml2-dev libxslt-dev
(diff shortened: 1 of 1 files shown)


aminvakil:
This reduces final image size almost 10%:
```
master/theharvester                   latest       c56a29c4e611   About a minute ago   397MB
quay.io/aminvakil/theharvester        latest       4a344b0c6f03   31 minutes ago       361MB
```

Also I guess the failed CI is because of change in branch name.

L1ghtn1ng:
Can you please fix the conflict

L1ghtn1ng:
To save you having to fix it will copy the changes in and merge it

L1ghtn1ng:
merged your changes in plus mine

aminvakil:
Thanks!

PR #773, opened 2021-06-17
Fix ModuleNotFoundError on install.

onrik:
Fix this problem on package install.
```bash
python setup.py install
Traceback (most recent call last):
  File "/home/onrik/Workspace/theHarvester-3.2.4/setup.py", line 2, in <module>
    from theHarvester.lib.core import Core
  File "/home/onrik/Workspace/theHarvester-3.2.4/theHarvester/lib/core.py", line 4, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```


L1ghtn1ng:
This PR will not fix that issue as you are just missing a library which can be fixed if you run the install for the dependencies

onrik:
But dependencies should installs due `setup.py install`

L1ghtn1ng:
What OS you using?

onrik:
Debian buster.

PR #799, opened 2021-07-13
improvement: bump Spyse API version & add pagination support

scanpatch:
Spyse will stop supporting API v3 soon, so I updated the integration to API v4. 

I also added pagination support so that now users can get up to 10,000 results with Spyse.

Please get in touch with me if you need a free Spyse API token to test the MR.


L1ghtn1ng:
Major thank you for this

PR #804, opened 2021-07-14
Update bufferoverun.py

JensTimmerman:
FDNS_A is sometimes None, then this function fails with a Nontype is not iterable exception.

### DIFF
--- a/theHarvester/discovery/bufferoverun.py
+++ b/theHarvester/discovery/bufferoverun.py
@@ -15,10 +15,11 @@ class SearchBufferover:
         responses = responses[0]
         dct = responses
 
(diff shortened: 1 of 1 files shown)


L1ghtn1ng:
Thanks for this, are you able to update the unit test to take this in to account?, sorry for the delay

JensTimmerman:
hi @L1ghtn1ng do you want me to add new tests? Or fix the failing test? I believe the failing test is because of an api not responding?

L1ghtn1ng:
Add a new test for this issue that this PR fixes and it runs fun locally the test they are probably blocking GitHub so do not worry about that

PR #821, opened 2021-08-02
New Dorks Added

Sbajary:
New google Dorks added.


NotoriousRebel:
The more I think about it, I think it would be more fruitful to revamp the google dorks to focus on email harvesting. Would you mind placing these dorks at the very top of the file and deleting everything else below. I can add a few more afterwards and push to master branch.

Sbajary (commented):
the unrelated dorks into email harvesting were deleted.

NotoriousRebel:
Looks good, thanks for PR

PR #910, opened 2021-11-09
company_domain seems more in line with the intention of -d

pierce403:
According to the rocketreach API documentation, `company_website_url`, which is the search term currently being used, requires the entire URL, including scheme etc. This will give back zero results with a search like "company.com". Alternatively, `company_domain` seems to give much nicer results in line with the way other discovery modules query for data.

https://rocketreach.co/api?section=api_section_ws_search


NotoriousRebel:
Thank you for the correction, merging!

PR #912, opened 2021-11-11
100x improvement in rocketreach API calls via pagination tweaks

pierce403:
Pagination was previously broken, requesting pages of 10, but incrementing the start value by 1 at each call. This change follows pagination suggestions as returned by the `pagination` field that gets returned.

The default page size is also set to 10 for some reason, and can be set up to 100, so by doing that there's also another 10x reduction in API calls which comes back faster and counts less against your daily rocketreach limit.

### DIFF
--- a/theHarvester/discovery/rocketreach.py
+++ b/theHarvester/discovery/rocketreach.py
@@ -25,8 +25,9 @@ class SearchRocketReach:
                 'User-Agent': Core.get_user_agent()
             }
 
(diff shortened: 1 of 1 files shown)


pierce403:
Some of the CI checks failed when I first sent this in, but it should be cleaned up now 👍

L1ghtn1ng:
Thanks for this

Issue #976, opened 2022-01-02
twitter source not working 


brokoder:
issue 1:
steps to reproduce:
python3 theHarvester.py -d yale.edu -b twitter 
## actual result
[*] Searching Twitter. 

[*] No Twitter users found.

[*] No IPs found.

[*] No emails found.

[*] No hosts found.

## expected result

[*] Twitter Users found: 72
---------------------
@ActusDei
@BharadwajSpeaks
@CCHYale
@Chalkemort
@CraigMCrews
@DrOnyemaOgbuagu
@DwightHallYale
@FedEx
@Goodman_lab
@ISPSYale
@JayArguello
@LUCIDM0NST3R
@Lin_YiZhang
@MaibachEd
@MikeWishnie
@NAChristakis
@NIFTYGYAN
@NSF
@Newsweek
@PMenonIFPRI
@SBS_Yale
@SarahLin5
@SarahRyanLowe
@SoftPresNetwork
@TataCompanies
@VBKerry
@Wikimedia
@WorldFellows
@Yale
@YaleAlumni
@YaleArch
@YaleBritishArt
@YaleDeptEHS
@YaleDivSchool
@YaleEGC
@YaleEconomics
@YaleEmotion
@YaleEnvironment
@YaleFaculty
@YaleISP
@YaleInsights
@YaleLawSch
@YaleMacMillan
@YaleMed
@YaleMusic
@YaleNursing
@YaleOCR
@YalePediatrics
@YalePsychology
@YaleSEAS
@YaleSOM
@YaleSPH
@akapczynski
@bekinddreambig
@catlenihan
@coursera
@gmail
@jasonlschwartz
@lauriesantos
@obermax2511
@sapphicmixer
@sicb
@soderej
@theYCCtweets
@tweetsbymidge
@vedatmilor
@yale
@yale_anthro
@yale_history
@yalelibrary
@yalemusic
@yarrowdunham

[*] No IPs found.

[*] No emails found.

[*] No hosts found.

issue 2:
adding proxy has no effect

system :
macOS v.12.1
python version 3.8.8

L1ghtn1ng:
Closing this out now that the your PR has been merged

PR #977, opened 2022-01-02
fix: fix issue with twitter source search and proxy varible not updat…

brokoder:
twitter source not working #976


L1ghtn1ng:
Thanks for this

PR #981, opened 2022-01-06
Package incompatible

rakibulinux:
Package incompatible requirements/base.txt
ERROR: spyse-python 2.2.3 has requirement requests~=2.26.0, but you'll have requests 2.27.1 which is incompatible.
ERROR: responses 0.13.4 has requirement urllib3>=1.25.10, but you'll have urllib3 1.25.8 which is incompatible.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -13,7 +13,8 @@ netaddr==0.8.0
 ujson==5.1.0
 pyppeteer==0.2.6
 PyYAML==6.0
-requests==2.27.1
+requests==2.26.0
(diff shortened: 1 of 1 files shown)


L1ghtn1ng:
This is a spyse issue so will need them to do a new release which will fix this issue but technically the only thing that needs changing is requests as urllib3 is as sub dependency of requests

L1ghtn1ng:
I have created a PR upstream https://github.com/spyse-com/spyse-python/pull/8. Sorry this has taken longer than expected to do, been dealing with long covid

PR #1180, opened 2022-08-13
BeVigil Source Added

alt-glitch:
BeVigil scans mobile applications and makes OSINT data available for security researchers.
Let me know if the code can be improved a bit :))

### DIFF
--- a/theHarvester/discovery/bevigil.py
+++ b/theHarvester/discovery/bevigil.py
@@ -1,0 +1,37 @@
+from theHarvester.lib.core import *
+
+class SearchBeVigil:
+    
+    def __init__(self, word):
+        self.word = word
+        self.totalhosts = set()
+        self.interestingurls = set()
+        self.key = Core.bevigil_key()
+        self.proxy = False
+
+    async def do_search(self):
+        subdomainEndpoint = f"https://osint.bevigil.com/api/{self.word}/subdomains/"
+        urlEndpoint = f"https://osint.bevigil.com/api/{self.word}/urls/"
+        headers = {
+            'X-Access-Token': self.key
+        }
+
+        responses = await AsyncFetcher.fetch_all([subdomainEndpoint], json=True, proxy=self.proxy, headers=headers)
+        response = responses[0]
+        for subdomain in response["subdomains"]:
+            self.totalhosts.add(subdomain)
+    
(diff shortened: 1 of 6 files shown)


L1ghtn1ng:
@alt-glitch Thank your for your request, you have included your api key so that needs removing and your will want to revoke that key as its now compromised due to the accidental inclusion of it. You will want to remove it and see if if you can remove it from git history. Unfortunately I do not know how to do that so cannot help on that one.

theHarvester/discovery/bevigil.py
```diff
@@ -0,0 +1,37 @@
...
+
+    async def get_hostnames(self) -> set:
+        return self.totalhosts
+
+    async def get_interestingurls(self) -> set:
+        return self.interestingurls
+
+    async def process(self, proxy=False):
+        self.proxy = proxy
+        await self.do_search()
```
L1ghtn1ng, line 37:
```suggestion
        await self.do_search()
        
```
This makes it compliant with flake8

alt-glitch:
Hi
Yes I think I mistakenly saw the `api_keys.yaml` file in the `.gitignore` so thought it wouldn't get added. I realised my mistake and have revoked the API key. Thank you for the heads up!

theHarvester/discovery/bevigil.py
```diff
@@ -0,0 +1,37 @@
...
+        self.interestingurls = set()
+        self.key = Core.bevigil_key()
+        self.proxy = False
+
+    async def do_search(self):
+        subdomainEndpoint = f"https://osint.bevigil.com/api/{self.word}/subdomains/"
+        urlEndpoint = f"https://osint.bevigil.com/api/{self.word}/urls/"
+        headers = {
+            'X-Access-Token': self.key
+        }
```
L1ghtn1ng, line 17:
```suggestion
        headers = {'X-Access-Token': self.key}
```
Just to remove un-needed space being taken up

L1ghtn1ng (changes requested):
Only a few things that need slightly changing but looks good :+1:

theHarvester/discovery/bevigil.py
alt-glitch: This seems to be the same? Am i missing something ? :)

L1ghtn1ng:
Thank you for adding this :+1:

PR #1284, opened 2022-12-16
Zoomeye asns format alignment

frapava98:
The error 
`'<' not supported between instances of 'int' and 'str'`
was raised when the program tried to sort the asns (line 510 of __main__.py: `total_asns = list(sorted(set(total_asns)))`) since zoomeye saved an int instead of a string.


L1ghtn1ng:
Thanks for this

PR #1287, opened 2022-12-19
Fix yahoo

frapava98:
The scan with yahoo was not going due to the process function signature not matching with the call at __main__.py line 121.


L1ghtn1ng:
Thank you for this

PR #1351, opened 2023-03-08
Trimmed the newline character from the domain input

Reelix:
Fixed a bug where piped inputs read from files would return incorrect results from certain data sources.


L1ghtn1ng:
Thanks for this

PR #1381, opened 2023-04-07
Added 4 new sources, Removed 2 Sources, Added custom DNS resolver, backported to python3.9, and other misc changes

NotoriousRebel:
#### Added 4 new sources:

- [Brave Search Engine](https://search.brave.com)
- [CriminalIP](https://www.criminalip.io/) (Requires API key) requested by some user in issue I can't find ATM but it's there
- [HunterHow](https://hunter.how/) (Requires API key)
- [Subdomainfinderc99](https://subdomainfinder.c99.nl/)

#### Removed 2 sources:

- Sublist3r (Website no longer exists and is no longer needed if it ever comes back up as it would be redundant)
- Threatcrowd (Website no longer exists)

#### Added --dns-resolve/ -r flag 

Can be used in one of four ways -r  or -r 1.1.1.1 or -r 1.1.1.1, 8.8.8.8  or -r resolvers.txt
By default subdomains are no longer automatically resolved which reduces overall runtime and should be a user choice if they want to use other resolving tools such as [MassDNS](https://github.com/blechschmidt/massdns). If they want to have it resolved the user must pass the -r flag in the corect format as aforementioned.

#### Backported to allow for users to use Python3.9 if they so choose
 Have not tested but Python3.8 might work if you change [this from a 9 to an 8](https://github.com/laramies/theHarvester/blob/master/bin/theHarvester#L7) and [this line](https://github.com/laramies/theHarvester/blob/master/theHarvester.py#L7)

#### Updated user agent list


L1ghtn1ng:
Hey Matt, can you please, rebase this PR as there are merge conflicts. Also the new modules are missing type checking on them which you can get pyre to automatically add them and its part of the dev.txt requirements and can you fix the flake8 errors and remove any unneeded commented out code as well please?

L1ghtn1ng:
oh and would be good if you could take a look at the mypy errors as well

L1ghtn1ng:
Oh and there is missing updates to the readme/wiki that will need including with this PR

PR #1417, opened 2023-05-10
Fix for bingapi

noaHson86:
hello,
i just did some small fixes on bingapi as it was not working at all

-changed api-endpoint
-fixed silent error (concat of string and list), no idea if this is the desired output now 
-fixed logic error in def process(...), do_search_api wasn't reachable at all


L1ghtn1ng:
Major thanks for this ☺️

PR #1460, opened 2023-07-02
Fix a bug about string splicing

captain686:
Fix the error "An exception has occurred: can only concatenate str (not "NoneType") to str" when the bevigil api lacks the key

### DIFF
--- a/theHarvester/discovery/bevigil.py
+++ b/theHarvester/discovery/bevigil.py
@@ -11,14 +11,13 @@ class SearchBeVigil:
         self.interestingurls: Set = set()
         self.key = Core.bevigil_key()
         if self.key is None:
+            self.key = ""
             raise MissingKey('bevigil')
         self.proxy = False
 
     async def do_search(self) -> None:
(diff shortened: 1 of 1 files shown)


L1ghtn1ng:
Thanks for this, can you have a look at the metals module please and copy that for the api key if statement, then will merge it in☺️ much appreciated this fix

captain686:
> 谢谢你，您能看一下 Metals 模块吗？并复制它作为 api key if 语句，然后将其合并到☺️非常感谢这个修复

Sorry, I don't understand what you mean by metal modules? Could you please explain it to me, thank you very much😊

captain686:
> Thanks for this, can you have a look at the metals module please and copy that for the api key if statement, then will merge it in☺️ much appreciated this fix

Sorry, I don't understand what you mean by metals module? Could you please explain it to me, thank you very much😊

L1ghtn1ng:
Apologies, autocorrect changed it, this is the module I was on about netlas.py

captain686:
Is there anything else I need to do?😀

L1ghtn1ng:
Yeah copy the if statement from that module as that is what it is meant to be for missing api keys

PR #1479, opened 2023-07-20
Add tomba module

benemohamed:
Add Tomba search engine


L1ghtn1ng:
The api key calling can be simplified to be the same as how censys is laid out if you could change it to that please

L1ghtn1ng:
Thank you for this module ☺️

Issue #1491, opened 2023-07-25
Errors when building docker image 


ibmchas:
## Note we do not support installing theHarvester on android

**Describe the bug**
Using Docker CE
```
Client: Docker Engine - Community
 Version:           24.0.4
 API version:       1.43
 Go version:        go1.20.5
 Git commit:        3713ee1eea
 Built:             Fri Jun 30 17:55:44 2023
 OS/Arch:           darwin/arm64
```
Running `docker build -t theharvester .` throws the following error during the build process 
```
Failed to build cffi
ERROR: Could not build wheels for cffi, which is required to install pyproject.toml-based projects
```

**To Reproduce**
Steps to reproduce the behaviour:
1. Following instructions - [installation](https://github.com/laramies/theHarvester/wiki/Installation#option-3---docker)
2. Docker build fails with the following error 
```
  Building wheel for cffi (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error

  × Building wheel for cffi (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [58 lines of output]
      /usr/lib/gcc/aarch64-alpine-linux-musl/12.2.1/../../../../aarch64-alpine-linux-musl/bin/ld: cannot find Scrt1.o: No such file or directory
      /usr/lib/gcc/aarch64-alpine-linux-musl/12.2.1/../../../../aarch64-alpine-linux-musl/bin/ld: cannot find crti.o: No such file or directory
      /usr/lib/gcc/aarch64-alpine-linux-musl/12.2.1/../../../../aarch64-alpine-linux-musl/bin/ld: cannot find -lssp_nonshared: No such file or directory
      collect2: error: ld returned 1 exit status
      Note: will not use '__sync_synchronize()' in the C code
      ***** The above error message can be safely ignored.

      /tmp/pip-build-env-h6xkgiir/overlay/lib/python3.11/site-packages/setuptools/config/setupcfg.py:293: _DeprecatedConfig: Deprecated config in `setup.cfg`
      !!

              ********************************************************************************
              The license_file parameter is deprecated, use license_files instead.

              By 2023-Oct-30, you need to update your project and remove deprecated calls
              or your builds will no longer be supported.

              See https://setuptools.pypa.io/en/latest/userguide/declarative_config.html for details.
              ********************************************************************************

      !!
```

**Expected behaviour**
Produce a working image.

**Screenshots**
N/A

**System Information (System that tool is running on):**
[...]

PR #1505, opened 2023-08-04
build: migrate to pep517 backend

### LINKED ISSUE #1435: [Feature Request] Support for pipx
**Feature Request or Bug or Another**
Feature Request

**Describe the feature request or bug or other**
Currently theHarvester does not play well with pipx which uses setuptools `entry_points` for installation. Setuptools themselves recommend moving dependencies to setup.cfg or pyproject.toml file as described in https://setuptools.pypa.io/en/latest/userguide/quickstart.html. This also refers to #1234.

**To Reproduce**
```
[2023-05-23 08:45 @ ~]$ pipx install git+https://github.com/laramies/theHarvester
  installed package theHarvester 4.3.0, installed using Python 3.9.2
  These apps are now globally available
    - restfulHarvest
    - theHarvester
done! ✨ 🌟 ✨
[2023-05-23 08:46 @ ~]$ theHarvester -d google.com
Traceback (most recent call last):
  File "/home/user/.local/bin/theHarvester", line 5, in <module>
    from theHarvester import __main__
  File "/home/user/.local/pipx/venvs/theharvester/lib/python3.9/site-packages/theHarvester/__main__.py", line 4, in <module>
    from theHarvester.discovery import dnssearch, takeover, shodansearch
  File "/home/user/.local/pipx/venvs/theharvester/lib/python3.9/site-packages/theHarvester/discovery/dnssearch.py", line 14, in <module>
    from aiodns import DNSResolver
ModuleNotFoundError: No module named 'aiodns'
```

**Expected behaviour**
The program should not have module errors after install with pipx.

**Screenshots**
---

**System Information (System that tool is running on):**
 - installed package theHarvester 4.3.0, installed using Python 3.9.2
 - pipx 1.2.0
 - Parrot OS 5.3

**Additional context**
---

branchvincent:
Fixes #1435

This migrates to `pyproject.toml` so the package can be install via `pip` (and thus `pipx`).
It partially handles `data_files` by installing `wordlists` with the package, ~but I haven't touched `api-keys.yaml` / `proxies.yaml` (these seem like user config files that don't need to be packaged/installed?)~ EDIT: now updated to install/read from `~/.theHarvester`

Demo:

```
pipx install git+https://github.com/branchvincent/theHarvester.git@pep-517
theHarvester --help
```


theHarvester/restfulHarvest.py
```diff
@@ -0,0 +1,46 @@
+import argparse
```
branchvincent, line 1: git was not able to track this, but this was just copied from the top-level `restfulHarvest.py`. This was needed since the [`[project.scripts]`](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#entry-points) callable needs to be an importable from the distributed package (i.e. a file under `theHarvester/`)

Note I've kept the existing top-level `restfulHarvest.py` functional by just importing from here

theHarvester/theHarvester.py
```diff
@@ -0,0 +1,25 @@
+import asyncio
```
branchvincent, line 1: same here, this is just relocating the top-level `theHarvester.py` (same reasoning as above)

theHarvester/lib/version.py
```diff
@@ -1,5 +1,7 @@
 # coding=utf-8
 
+VERSION = "4.4.3"
```
branchvincent, line 3: This was needed for setuptools to discover the dynamic [`version`](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) (it only supports a static constant, not a callable function)

pyproject.toml
```diff
@@ -1,7 +1,41 @@
...
+]
+dynamic = ["dependencies", "optional-dependencies", "version"]
+
+[project.scripts]
+theHarvester = "theHarvester.theHarvester:main"
+restfulHarvest = "theHarvester.restfulHarvest:main"
+
+[tool.setuptools.dynamic]
+version = { attr = "theHarvester.lib.version.VERSION" }
+dependencies = { file = "requirements/base.txt" }
```
branchvincent, line 26: Note I excluded this to keep my changes as minimal as possible, but we could also include the `requirements/dev.txt` as a `dev` extra:

```toml
optional-dependencies.dev = { file = "requirements/dev.txt" }
```

which would allow developers to install using the common `pip install -e '.[dev]'`. However, it doesn't support pip flags, like https://github.com/laramies/theHarvester/blob/f661afc2559eaba5daf15e36d7423672e1091a9b/requirements/dev.txt#L1
That line wouldn't be needed anymore since `pip` would already install the base requirements, but it would require a different workflow for devs as well as updating the wiki

L1ghtn1ng:
Okay, am happy for that change to go in, also can you add me and matt
to the authors section please

NotoriousRebel:
What happened to the Dockerfile changes? It should be trivial to update the file to make it work correctly [as the solution] was posted a few weeks ago. May be worthwhile adding it to the docker registry once completed and thoroughly tested.

L1ghtn1ng:
Wrong topic Matt but I added the changes that you put in the issue

On 26 Aug 2023, at 23:46, Matt ***@***.***> wrote:

﻿

What happened to the Dockerfile changes? It should be trivial to update the file to make it work correctly [as the solution] was posted a few weeks ago. May be worthwhile adding it to the docker registry once completed and thoroughly tested.

branchvincent:
> Okay, am happy for that change to go in, also can you add me and matt to the authors section please

Done! I also pushed a small tweak to include the data files in `theHarvester.data`

> What happened to the Dockerfile changes

Oops, I tried to minimize my changes here so it was easier for review, but didn't mean to revert that. I've pushed an update to make it compatible. Note there was previously two `ENTRYPOINT`s but Docker only [supports one](https://docs.docker.com/engine/reference/builder/#entrypoint)

L1ghtn1ng:
Ah sorry Matt did not see there was any docker stuff, my bad lol

On 27 Aug 2023, at 01:44, Branch Vincent ***@***.***> wrote:

﻿

Okay, am happy for that change to go in, also can you add me and matt to the authors section please

Done! I also pushed a small tweak to include the data files in theHarvester.data

What happened to the Dockerfile changes

Oops, I tried to minimize my changes here so it was easier for review, but didn't mean to revert that. I've pushed an update to make it compatible. Note there was previously two ENTRYPOINTs but Docker only supports one<https://docs.docker.com/engine/reference/builder/#entrypoint>

NotoriousRebel:
All good also realized I should not have typed that on mobile yesterday, the solution I was referencing to was a [previous issue](https://github.com/laramies/theHarvester/issues/1491). We may just need to add those `RUN` commands or the docker container may not work.

L1ghtn1ng:
This needs rebasing

branchvincent:
@L1ghtn1ng done

L1ghtn1ng:
can you check please to make sure stash.py does not get broken by this?

L1ghtn1ng:
also if you are able to write some tests for these changes that would also be awesome

branchvincent:
> can you check please to make sure stash.py does not get broken by this?

Yea how would I do that? Note I'm not a user of theHarvester, just wanted to help with packaging

> also if you are able to write some tests for these changes that would also be awesome

Sure, since the only logic changes are related to loading config files, I could write a test exercising that?

And beyond the scope of this PR, but an easy way to catch packaging-related issues is to move to a `src` layout: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/. That way, your test suite can actually test your installed package (which would catch missing data files and #1516 for example).

L1ghtn1ng:
> Yea how would I do that? Note I'm not a user of theHarvester, just wanted to help with packaging

This line really https://github.com/laramies/theHarvester/blob/c397275cac66a448f9b2c9386df215a68a471b67/theHarvester/lib/stash.py#L8 

> sure, since the only logic changes are related to loading config files, I could write a test exercising that?

That would be great :-)

branchvincent:
@L1ghtn1ng added a test, and afaict `stash.py` works fine

PR #1670, opened 2024-02-24
Fix ip-address parsing for IPv6 addresses

jeffguy:
Sometimes, the host string uses IPv6 instead of IPv4, so the address has colons. This can result in a string like

`google.com:2607:f8b0:4006:80b::200e`

If we then run

`subdomain, addr = host.split(":")`

`host.split(":")` returns a list of seven strings because their are six colons. Setting `(subdomain, addr)` to a list of seven strings doesn't work (`ValueError: too many values to unpack (expected 2)`). What we really want here is

`subdomain, addr = host.split(":", 1)`

which splits only on the first colon, so a list of two strings is returned, and the whole IPv6 address gets assigned to the `addr` variable.

### DIFF
--- a/theHarvester/__main__.py
+++ b/theHarvester/__main__.py
@@ -1057,7 +1057,7 @@ async def start(rest_args: Optional[argparse.Namespace] = None):
             for host in full:
                 if ":" in host:
                     # TODO parse addresses and sort them as they are IPs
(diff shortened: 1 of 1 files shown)


L1ghtn1ng:
Thank you for submitting this, much appreciated ☺️

PR #1760, opened 2024-06-17
Properly clean up apt footprint in Dockerfile

PeterDaveHello:
According to the official best practices guidelines outlined in the following document: https://docs.docker.com/develop/develop-images/instructions/#apt-get

The official Debian/Ubuntu images have been configured to do `apt-get` clean automatically. Consequently, performing `apt clean` is unnecessary. As no package has been removed, another `apt autoremove` is also unnecessary. Instead, it'll be helpful to clear the cache for the index file by executing this command: `rm -rf /var/lib/apt/lists/*`.


L1ghtn1ng:
Thanks ☺️

PR #1900, opened 2024-12-13
Change: methods of searching subdomains in SearchNetlas

thurrsense:
The current implementation of theHarvester uses an outdated data acquisition method Netlas.io. This leads to inaccurate data being received.
In this PR, I:
- Changed the do_search() function by adding a POST request to "app.netlas.io/api/domains/download ", this allows you to get all the existing ones in the database Netlas.io subdomains for a specific domain.
- Added the do_count() function, which pre-calculates the number of existing subdomains using a GET request.
- The ability to limit the number of domains received via the "--limit" parameter has also been added.
The fix was tested under the following conditions: locally with test data.

### DIFF
--- a/theHarvester/__main__.py
+++ b/theHarvester/__main__.py
@@ -604,7 +604,7 @@ async def start(rest_args: argparse.Namespace | None = None):
 
                 elif engineitem == 'netlas':
                     try:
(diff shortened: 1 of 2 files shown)


L1ghtn1ng:
Thanks for this

PR #1976, opened 2025-04-22
Update Dockerfile

noarche:
Dockerfile update to Successfully install on server running docker io


L1ghtn1ng:
If we use uv that can can install the required version of Python, would like to use that if possible? If not can we have Python 3.12.x installed please☺️

PR #1998, opened 2025-05-18
Added a module for searching subdomains via the dnsdumpster service API

p3test:
Added a module for searching subdomains via the dnsdumpster service API


L1ghtn1ng:
Thanks for this

PR #2059, opened 2025-07-08
point at docker port

digininja:
Docker is listening on 8080 by default so pointing the example at that rather than 5000 seems to make more sense.

### DIFF
--- a/theHarvester/lib/api/api_example.py
+++ b/theHarvester/lib/api/api_example.py
@@ -33,7 +33,7 @@ async def main() -> None:
     Just a simple example of how to interact with the rest api
     you can use requests instead of aiohttp or whatever you best see fit
     """
-    url = 'http://127.0.0.1:5000'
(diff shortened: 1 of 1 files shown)


L1ghtn1ng:
Yeah makes sense to move docker to be port 5000

On 8 Jul 2025, at 09:54, Robin Wood ***@***.***> wrote:

﻿
[https://avatars.githubusercontent.com/u/117081?s=20&v=4]digininja left a comment (laramies/theHarvester#2059)<https://github.com/laramies/theHarvester/pull/2059#issuecomment-3047972466>

would it be worth moving docker to 5000 as well so they are in sync?

Or at least documenting in some way that you need to change it for docker?

L1ghtn1ng:
fixed this in master https://github.com/laramies/theHarvester/commit/833b062649b7e7f77e075a81188bcbdcff06a54a

Issue #1371, opened 2023-03-28
how to add shodan api on version 4.2.0


stevendemu:
How can i add mu shodan api to theharvester

i have looked online but cant find any maual

**Feature Request or Bug or Other**
Feature Request | Bug | Other

**Describe the feature request or bug or other**
A clear and concise description of what the bug, feature request,
or other request is.

**To Reproduce**
Steps to reproduce the behaviour:
1. Run tool like this: '...'
2. See error

**Expected behaviour**

**Screenshots**
If possible please add screenshots to help explain your problem.

**System Information (System that tool is running on):**
 - OS: [e.g. Windows10]
 - Version [e.g. 2.7]

**Additional context**

NotoriousRebel:
Adding shodan to the api-keys.yaml is all you need to do they may have changed how their API is strucuted as they severely reduced total number of credits like a year ago. Shodan will be able to get subdomains and I guess provide more info once I get a chance to see what is breaking if anything. So let me know as you should just be able to do `python3 theHarvester.py -d uber.com -b bing -l 200 --dns-resolve 1.1.1.1,8.8.8.8 --shodan`

h3b4r1:
Shodan seems broken at the moment.
Getting the following error.

[!] An error occurred with Shodan: 'str' object has no attribute 'items'

NotoriousRebel:
I will be fixing this soon and allowing for shodan to be used as a normal module to query for domains and possibly revise the scanning IPs found. Keep in mind that Shodan has changed their API rate limits and severely reduced credits used for querying IPs which may mean if that module was fixed using theHarvester just once could use up every query credit depending on how many IPs were found. More information will be put in docstrings, the wiki, or the readme, once this is further investigated/clarified.

leebaird:
See PR #2069.

L1ghtn1ng:
fixed by #2069 that has now been merged

Issue #875, opened 2021-10-05
[Bug] shodan module broken


leebaird:
Bug

The shodan module no longer returns any results.
Tried both the Kali and GitHub versions.
No errors shown.

To Reproduce
Kali: theHarvester -d yale.edu -s
Added API key to /etc/theHarvester/api-keys.yaml

GitHub: ./theHarvester.py -d yale.edu -s
Added API key to api-keys.yaml

Expected Behavior
15 results

System Information
OS: Kali Linux
Version 2021.3

Additional Information
I just tested the built in shodan tool in Kali with my API and it works fine.

py2n:
i have found the problem.
in the main __main__.py there is a shodan section which uses api to fetch required data.
there is a for loop which search for ip addresses in shodan.io. if the IP address is not present in shodan Database theHarvester crashes and it does not continue for searching other IP addresses.
i have solved the problem by surrounding the api call with try/except block which resolved the bug.
` # Shodan
    shodanres = []
    if shodan is True:
        import json
        print('\033[94m[*] Searching Shodan. \033[0m')
        try:
            for ip in host_ip:
                try:
                    print(('\tSearching for ' + ip))
                    shodan = shodansearch.SearchShodan()
                    shodandict = await shodan.search_ip(ip)
                    await asyncio.sleep(2)
                    rowdata = []
                    for key, value in shodandict[ip].items():
                        if str(value) == 'Not in Shodan' or 'Error occurred in the Shodan IP search module' in str(value):
                            break
                        if isinstance(value, int):
                            value = str(value)

                        if isinstance(value, list):
                            value = ', '.join(map(str, value))
                        rowdata.append(value)
                    shodanres.append(rowdata)
                    print(json.dumps(shodandict[ip], indent=4, sort_keys=True))
                    print('\n')
                except Exception as e:
                    print('[SHODAN-error] {e}')
        except Exception as e:
            print(f'\033[93m[!] An error occurred with Shodan: {e} \033[0m')
`

py2n:
@L1ghtn1ng could you please check this?
if it's OK then i create a pull request?

cstyrsyn:
I was having the same issue and the above fix appears to work in allowing the module to continue after not finding an IP in shodan.

To Reproduce
theHarvester -d yale.edu -b censys -s  
(I used censys as just an example I could hit a few times for testing)

System Information
theHarvester 4.0.3
OS: Kali Linux
Version 2022.1

Error output
[...]

PR #2069, opened 2025-07-12
Fix Shodan

ibrahmsql:
Fix Shodan module issues and convert to regular module #875  #1371


ibrahmsql:
Fixed removed the duplicate return  thanks

PR #2097, opened 2025-08-05
Add workflow to publish to PyPI

### LINKED ISSUE #1501: PyPI package
Is there a way we can push theHarvester on PyPI to ease up the installation on automated environments ? A `pip install theHarvester` would be nice as an easy install method.

branchv:
Fixes #1501

This adds a simple workflow to publish to PyPI on any new git tag. You just need to register this workflow with PyPI [here](https://pypi.org/manage/account/publishing/) (as explained in [this python guide](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#configuring-trusted-publishing)).

Note [theHarvester already exists](https://pypi.org/project/theHarvester/) on PyPI, so hopefully you don't have any trouble gaining access

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,19 @@
+name: Release
+
+on:
+  push:
+    tags: ["*"]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/theHarvester
+    permissions:
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v6
+      - run: uv build
+      - run: uv publish


L1ghtn1ng:
https://github.com/pypi/support/issues/9189

munzzyy:
Nice work here, this is the right shape for a publish workflow. Using `uv publish` with an `environment: pypi` block means you're on trusted publishing (OIDC) rather than a long-lived `PYPI_API_TOKEN` secret, and `id-token: write` is scoped to just the publish job instead of handed out repo-wide. That's the hard part and you got it right.

Two things I'd tighten up before merge, both because they'd bring this file in line with what the rest of `.github/workflows/` in this repo already does:

1. **Pin the actions to a commit SHA, not a floating tag.** Every other workflow here (`theHarvester.yml`, `docker-build-push.yml`, `scorecards.yml`) pins `actions/checkout` and friends to a full SHA with the version as a trailing comment, e.g. `actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0`. This workflow uses `actions/checkout@v4` and `astral-sh/setup-uv@v6`, which are mutable tags — anyone who compromises those actions' repos (or the tag itself gets moved) gets code execution in the one job that has a live PyPI credential attached to it. That's the highest-value job in the whole CI to have a floating dependency in, so it's worth being at least as strict here as the rest of the repo already is.

2. **Add `contents: read` explicitly alongside `id-token: write`.** Right now only `id-token: write` is declared. It'll probably still work depending on the repo's default token permissions setting, but relying on the implicit default is fragile and it's one line to make it explicit and self-documenting.

Small optional one: the other workflows all run `step-security/harden-runner` with an egress audit as the first step. Might be worth adding here too, mostly for the outbound-call visibility on a job that's about to authenticate to PyPI.

With those two fixed I think this is good to go:

```yaml
jobs:
  publish:
    runs-on: ubuntu-latest
    environment:
      name: pypi
      url: https://pypi.org/p/theHarvester
    permissions:
      id-token: write
      contents: read
    steps:
      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
      - uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
      - run: uv build
      - run: uv publish
```

Trigger (`push: tags`) and job isolation both look fine as-is. Mostly this just needs a maintainer to register the trusted publisher on PyPI and merge it — it's been sitting here almost a year and the core approach is already solid.

L1ghtn1ng:
Yeah I know it's been here for that long, but I'm trying to get the namespace on pypi so I can merge it as we do not own the namespace.

Issue #1508, opened 2023-08-04
Replace `except Exception`s with proper error handling


yoonthegoon:
<img width="263" alt="image" src="https://github.com/laramies/theHarvester/assets/71526721/a20911a7-5495-4e5b-be7b-ebb5b19edb60">

Catching all errors with `Exception` is enticing, but has several issues.
Properly catching the errors you expect to get: 

1. improves readability and maintainability
2. improves error diagnostics
3. avoids unexpected behavior

There are cases where you may be passing one argument to a function that requires two, like `filter` that will go unnoticed while running or testing, as the error it raises just gets caught by a bare `Exception`.
I want to avoid situations like this; exceptions should catch only expected behavior.

leebaird:
Love this idea.

theHexable:
**First**, There Another approach to solve this with the general exception, using the following code:
`print(f'An exception has occurred: {e.__class__.__name__} : {url}')`
will show us the exception type without specifying custom solution for every type

**Second**, The annoying thing about this, is there are many messages printed after exceptions caught.
see the following as sample:
```
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=640
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=870
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=180
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=90
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=610
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=110
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=710
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=440
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=420
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=140
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=950
An exception has occurred: ContentTypeError : https://sonar.omnisint.io/all/www.edge-security.com?page=350
[...]

PR #2120, opened 2025-08-30
Replace except Exceptions with proper error handeling

Rafaelorr:
This fixes this issue #1508


theHarvester/lib/core.py
```diff
@@ -554,3 +554,8 @@ async def fetch_all(
                     )
                 else:
                     return await asyncio.gather(*[AsyncFetcher.fetch(session, url, params, json) for url in urls])
+
+def show_default_error_message(engine_name :str, word :str, error) -> None:
```
L1ghtn1ng, line 558: This is not used anywhere?

Issue #2128, opened 2025-09-11
Add new API interfaces


l3m0n:
https://windvane.lichoin.com/docs-api

l3m0n:
[windvane api en documentation.pdf](https://github.com/user-attachments/files/22310185/windvane.api.en.documentation.pdf)

OK, I have translated it. See if this helps you.

ibrahmsql:
Hello @l3m0n I’ve added windvane support in #2131 PR  I've attached a screenshot in case you’d like to take a look
<img width="1143" height="897" alt="Image" src="https://github.com/user-attachments/assets/fbb65735-a527-4b5e-8411-485f2e97717b" />

l3m0n:
Thanks

Issue #2103, opened 2025-08-09
[Feature Request] Add new modules


leebaird:
Modules from Subfinder (https://github.com/projectdiscovery/subfinder) that are not included in theHarvester:

chaos
chinaz
commoncrawl
digitalyama
digitorus
dnsdb
dnsrepo
facebook
fofa
gitlab
leakix
pugrecon
quake
reconcloud
redhuntlabs
riddler
robtex
threatbook
threatcrowd
waybackarchive

PR #2131, opened 2025-09-15
Add 10 new discovery modules addressing issues #2103 and #2128

ibrahmsql:
I will fix the bugs and reopen the pr

### DIFF
--- a/theHarvester/data/api-keys.yaml
+++ b/theHarvester/data/api-keys.yaml
@@ -25,6 +25,10 @@ apikeys:
   dnsdumpster:
     key:
 
+  fofa:
+    key:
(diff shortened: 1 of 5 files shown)


L1ghtn1ng:
@ibrahmsql I would say as the Facebook ct api is going away this year probably not worth having it around unless people thing otherwise?

ibrahmsql:
I see your point thanks maybe we can keep it until it’s fully deprecated but I’m open to removing it if others think its not worth it

L1ghtn1ng:
I would say it's not worth it as we would just have to remove it again at a later date so more work if you see my point?

PR #2363, opened 2026-06-10
Remove deprecated gethostbyname in favor of getaddrinfo

Vizonex:
Revisiting this repo and being a maintainer of aiodns I found a function that was considered by our team as deprecated so I migrated the function away from gethostbyname in favor of getaddrinfo so that it doesn't need migrating to later.


L1ghtn1ng:
Thanks for this

PR #2404, opened 2026-07-13
Respect --limit for VirusTotal results

ShivammVishwakarma:
## Summary

This pull request fixes the VirusTotal backend so that it respects the user-supplied `--limit` (`-l`) option.

## Changes

- Pass the `limit` value from `__main__.py` to `SearchVirustotal`.
- Stop requesting additional VirusTotal pages once the requested number of unique hosts has been collected.
- Trim the final hostname list to the requested limit before returning it.

## Testing

Verified with:

```bash
theHarvester -d openai.com -b virustotal -l 5
```

The output now returns exactly 5 hosts, matching the requested limit.

### DIFF
--- a/theHarvester/__main__.py
+++ b/theHarvester/__main__.py
@@ -1241,7 +1241,7 @@ async def start(rest_args: argparse.Namespace | None = None):
 
                 elif engineitem == 'virustotal':
                     try:
-                        virustotal_search = virustotal.SearchVirustotal(word)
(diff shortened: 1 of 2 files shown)


NotoriousRebel:
Closing this as limit is really meant for the underlying APIs not to reduce total hostnames found. It has been a while since the API was looked at and will be rexamined before the next release in the very near future.

PR #2406, opened 2026-07-21
refactor: add structured verbose logging

NotoriousRebel:
## Summary

- add a single `--verbose` flag for informational diagnostics
- replace direct `print()` calls with operator-output or diagnostic loggers
- keep normal operator results on stdout and diagnostics on stderr
- preserve host-owned logging handlers when the package or REST app is embedded
- prevent raw provider responses and Dehashed credential rows from reaching logs
- enforce the no-`print()` boundary with Ruff T20

## Why

The project needs useful opt-in diagnostics without changing normal CLI output, taking over an embedding application's logging policy, or exposing sensitive provider data. This refactor establishes those boundaries across the codebase.

This is the upstream version of [NotoriousRebel/theHarvester#87](https://github.com/NotoriousRebel/theHarvester/pull/87), replayed onto `laramies/theHarvester:dev` as a clean focused branch.

## Validation

- 108 offline regression tests passed; live-network Cert Spotter, THC, and OTX tests were excluded
- Ruff check and format passed
- repository-wide Ruff T20 passed
- mypy passed
- final patch equivalence against the reviewed fork PR was confirmed with `git range-diff`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,7 +139,8 @@ select = ["E",
     "PT",
     "TC",
     "FURB",
-    "ASYNC"
+    "ASYNC",
+    "T20"
     ]
 ignore = [
     "B018",
--- a/tests/discovery/test_bitbucket.py
+++ b/tests/discovery/test_bitbucket.py
@@ -1,0 +1,23 @@
+import logging
+from unittest.mock import AsyncMock
+
+import pytest
+
+from theHarvester.discovery import bitbucket
+
+
+@pytest.mark.asyncio
+async def test_process_does_not_log_error_body(monkeypatch, caplog) -> None:
+    monkeypatch.setattr(bitbucket.Core, 'bitbucket_key', lambda: 'test-key')
+    search = bitbucket.SearchBitBucket('owner/repository', limit=10)
+    monkeypatch.setattr(
+        search,
+        'do_search',
+        AsyncMock(return_value=('', {'secret': 'provider-secret-payload'}, 500, {})),
+    )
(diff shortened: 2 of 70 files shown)


L1ghtn1ng:
I found two logging regressions that should be addressed before this is ready to merge:

1. The unexpected-error paths in `/sources`, `/dnsbrute`, and `/query` now call `logger.info(...)`. Under Uvicorn's default logging configuration, the `theHarvester.lib.api.api` logger inherits the root `WARNING` level, so these diagnostics are discarded. I reproduced this with Uvicorn's actual `LOGGING_CONFIG`: the `INFO` diagnostic was absent while an `ERROR` diagnostic was emitted. That leaves operators with a generic HTTP 500 and no corresponding server-side failure detail. Please use `logger.exception(...)` (or otherwise configure a dedicated application logger in the standalone REST entry point) and add a regression test using Uvicorn's logging configuration. `logger.exception(...)` also removes the need to build the traceback manually.

2. `configure_logging(verbose=True)` sets the `theHarvester` package logger to `INFO`, but a later `configure_logging(verbose=False)` leaves it at `INFO`. A `False -> True -> False` sequence produced effective levels `30 -> 20 -> 20`, so verbose diagnostics remain enabled in long-lived embedded/test processes. Please track and restore only logger state owned by this configuration helper, or use a dedicated handler/filter, and cover that sequence in a test.

There is also a direct merge conflict with #2408 in `discovery/bravesearch.py` and overlapping work in `lib/core.py`. When rebasing, the resolution needs to retain #2408's credential injection/lazy proxy loading as well as this PR's sanitized logger calls and T20 boundary.

PR #2408, opened 2026-07-21
Introduce lazy and injectable configuration boundaries

NotoriousRebel:
## Summary

- defer proxy-file loading until `AsyncFetcher` first needs proxies
- cache proxy configuration after first access
- add filesystem and in-memory credential adapters
- migrate Brave credential lookup to the injectable boundary
- preserve the public `Core.brave_key()` accessor as a compatibility wrapper
- document the boundary rationale and link Brave's first-party query API

## Why

Importing core code previously read proxy configuration immediately, and provider constructors reached directly into global filesystem-backed credential state. Those hidden reads made imports depend on an operator's machine and made isolated tests and embedded use harder.

Production lookup behavior and directory precedence remain unchanged. Brave is the single tracer implementation; this PR does not introduce a broader configuration framework.

This is the upstream version of [NotoriousRebel/theHarvester#84](https://github.com/NotoriousRebel/theHarvester/pull/84), rebased onto current `laramies/theHarvester:dev`.

## Composition

The rebase preserves #2406's merged logging behavior. The requested `Core.brave_key()` compatibility regression is included.

## Validation

- focused configuration and Brave tests: 13 passed
- offline remainder: 150 passed, 4 skipped, 2 live tests deselected
- Ruff check and format check passed
- mypy passed
- `git diff --check` passed
- Standards review: clean
- Spec review: clean

### DIFF
--- a/tests/lib/test_configuration_contract.py
+++ b/tests/lib/test_configuration_contract.py
@@ -1,0 +1,144 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, NamedTuple
+
+import pytest
+import yaml
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from pathlib import Path
+    from types import ModuleType
+
+    from theHarvester.lib.core import Core
+
+
+class ConfigurationEnvironment(NamedTuple):
+    core: type[Core]
+    module: ModuleType
+    directories: list[Path]
+
+
+@pytest.fixture
+def configuration_environment(
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path: Path,
+) -> ConfigurationEnvironment:
+    monkeypatch.setenv('HOME', str(tmp_path))
+    import theHarvester.lib.core as core_module
+
+    directories = [tmp_path / name for name in ('user', 'system', 'local')]
+    for directory in directories:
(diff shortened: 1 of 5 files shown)


L1ghtn1ng:
Please preserve the existing `Core.brave_key()` accessor. The `dev` base exposes this method, but this PR removes it even though the stated contract says production lookup behavior remains unchanged. Existing integrations or tests that call `Core.brave_key()` will fail with `AttributeError` after the merge.

Recommended change: retain `Core.brave_key()` as a compatibility wrapper (it can delegate through the filesystem credential adapter) and add a regression test asserting that the accessor remains available and returns the configured Brave key.

As a non-blocking design improvement, consider expressing `CredentialAdapter` as a `Protocol` with `get(provider, field)` rather than a union of the two bundled implementations. The current union makes the runtime boundary injectable but causes static type checkers to reject compatible third-party adapters.

This PR also conflicts directly with #2406 in `discovery/bravesearch.py`. I recommend landing this configuration boundary first, then rebasing #2406 so its logging conversion is applied to the adapter-based implementation without reintroducing direct `print()` calls.

L1ghtn1ng:
You all done on this as if so am happy for it to be merged

Issue #94, opened 2018-10-18
[Feature Request] Add intelx.io


tacticalDevC:
Intelx.io is a new search engine made in Austria/Czechia. It utilizes the dark web andnorth corean sites. It would be a HUGE improvement if you add this search engine!

NotoriousRebel:
This looks very interesting I could start working on this sometime soon, the only problem is that they are onion sites and the requests library doesn't work very well with them.

NotoriousRebel:
@tacticalDevC I believe the request library would be easiest as you can have an optional argument when sending a request to go through a proxy. Would using proxies off a free proxy list be sufficient?

tacticalDevC:
@NotoriousRebel I guess so

NotoriousRebel:
This will be added in the somewhat near future as their API page https://api.intelx.io/ leads to a 404 error, it will be added when better support for their API arrives.

Issue #2403, opened 2026-07-11
Baidu search module broken due to anti-bot / security verification (CAPTCHA)


maksimtech:
Description:

### Describe the bug
The `baidusearch` module is currently unable to fetch any results (IPs, emails, or hosts) because Baidu implements aggressive anti-bot protections and blocks requests coming from cloud environments/datacenters (tested on GitHub Codespaces). 

Instead of returning search results, Baidu returns a security verification/CAPTCHA page (`<title>百度安全验证</title>`), which results in a static response length and zero parsed items.

Additionally, the hardcoded `%40` (the `@` symbol) forced before the domain query parameter (`wd=%40{self.word}`) breaks compatibility with modern Baidu search syntax.

### steps to reproduce
1. Run theHarvester using Baidu as a source:
   ```bash
   uv run ./bin/theHarvester -d microsoft.com -l 100 -b baidu

    See the final output showing zero results:
    Plaintext

    [*] Searching Baidu. 
    [*] No IPs found.
    [*] No emails found.
    [*] No people found.
    [*] No hosts found.

Debugging & Technical Details

Dumping the raw HTML response from www.baidu.com/s?wd=... reveals that Baidu serves a Baidu Security Verification page (zh-CN):
HTML

<!DOCTYPE html>
<html lang="zh-CN">
<head>
    <meta charset="utf-8">
    <title>百度安全验证</title> <!-- Baidu Security Verification -->
    ...

Because the engine always encounters this challenge page, myparser.py fails to extract any meaningful intelligence.
Possible Solutions / Discussion

    Query string fix: At a minimum, the %40 prefix should be removed from baidusearch.py since it messes up current Baidu operators.

    Deprecation/Fix: Discuss whether the Baidu module needs to be deprecated/removed or if advanced anti-scraping bypasses (like custom cookies or specific residential proxy requirements) should be officially documented for this source.

NotoriousRebel:
Thank you for reporting, will look into this may need to use playwright with a headless browser instead of aiohttp.

thejasmy548-hub:
Lathesh

NotoriousRebel:
Closing this for now as this has been fixed with multiple PRs including #2429 and #2579. It still seems like captchas eventually become an issue but the module is no longer fully broken. A lot of other modules have been added along with ways to start benchmarking each source to see which sources truly yield unique subdomains so we will have a better picture soon into whether it's worth spending more time on trying to extract as many results from Baidu as possible.

[...]

PR #2411, opened 2026-07-22
Harden Python CI and isolate passive provider smoke tests

NotoriousRebel:
## Summary

- grant routine and manual workflows only `contents: read`
- keep routine Ruff checks non-mutating
- make ordinary pytest offline by default with a Python socket tripwire
- require both the `live_network` marker and `--run-live-network` for live tests
- replace routine Cert Spotter, OTX, and THC calls with synthetic canned responses
- move bounded passive-provider smoke checks to a separate manual-only workflow using `example.com`

## Why

Routine push and pull-request CI should be deterministic, read-only, and free of live reconnaissance. The previous workflow rearrangement still ran unmarked provider tests during ordinary pytest. This update enforces the separation in the test harness as well as the workflow.

The socket tripwire is defense in depth, not OS-level egress isolation: native resolvers, extensions, and subprocesses can bypass it. The workflow and test policy must not be described as fully hermetic.

This builds on [NotoriousRebel/theHarvester#93](https://github.com/NotoriousRebel/theHarvester/pull/93) and the offline harness reviewed in fork PR #94, rebased onto current `laramies/theHarvester:dev`.

## Validation

- focused provider contracts: 24 passed, 6 skipped
- harness and workflow contracts: 5 passed
- full suite: 143 passed, 6 skipped
- live collection: exactly 6 tests; missing marker/flag combinations are rejected
- Ruff check and format check passed
- mypy passed for product and changed tests
- `git diff --check` passed
- Standards review: clean
- Spec review: clean


L1ghtn1ng:
One issue is worth fixing before merge: routine push/PR CI still performs live third-party calls despite this PR moving the CLI smoke checks behind `workflow_dispatch`.

The workflow still runs `pytest tests/**`. In the successful PR job (run `29883866837`, Python 3.12), pytest collected `tests/discovery/test_otx.py` and `tests/discovery/test_thc.py`, and the Harden Runner audit recorded the pytest process contacting `otx.alienvault.com` and `ip.thc.org`. It also contacted `www.baidu.com` and `wappass.baidu.com`; `tests/test_security.py::test_error_responses_do_not_leak_internal_paths` calls `/query?domain=test&source=baidu`, which executes the real provider. This means every routine run still depends on live services and performs live queries three times across the Python matrix.

Please make the default pytest selection network-free: for example, mark the OTX/THC tests as integration tests and exclude that marker in routine CI (running it only in the manual smoke job), and mock the provider call in the API security test. The least-privilege and check-only Ruff changes otherwise look good.

L1ghtn1ng:
There is still one issue worth fixing before merge: routine pytest is not fully offline yet.

In the updated successful Python 3.12 job (run `30411209106`, job `90447470887`), the Harden Runner audit recorded `python3.12` connecting to `www.baidu.com` during `tests/test_security.py`. The unchanged `test_error_responses_do_not_leak_internal_paths` still requests `/query?domain=test&source=baidu`, which dispatches the real Baidu provider. That path bypasses the Python socket tripwire, so ordinary push/PR CI continues making a live provider request across the matrix.

Please mock the query execution boundary for this test (for example, patch `theHarvester.lib.api.api.__main__.start` to return a canned result or raise a controlled exception), or use a validation/error case that cannot dispatch a provider. A regression assertion should ensure this API test never reaches `AsyncFetcher.fetch_all`. The marked Cert Spotter/OTX/THC tests, socket guard, read-only workflow, and separate manual smoke workflow otherwise look good.

L1ghtn1ng:
then again basiu is broken so maybe we just remove it? see #2403 to see what I mean

PR #2437, opened 2026-07-30
fix: sync uv lock cutoff metadata

NotoriousRebel:
## Summary

- sync the committed uv lock with the configured seven-day dependency cutoff
- reuse existing upstream commit 424b672c
- unblock Docker builds that run uv sync --locked

## Verification

- uv lock --check --offline, repeated twice
- UV_PROJECT_ENVIRONMENT=<temporary path> uv sync --locked --no-dev --no-cache --compile-bytecode

Both commands passed locally. A full Docker build was blocked before the install layer by a registry metadata timeout while fetching ghcr.io/astral-sh/uv:latest.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -7,6 +7,10 @@ resolution-markers = [
     "python_full_version < '3.14'",
 ]
 
+[options]
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
+exclude-newer-span = "P7D"
+
 [[package]]
 name = "aiodns"
 version = "4.0.4"


L1ghtn1ng:
this is not needed, have fixed the docker build in dev already

NotoriousRebel:
Thanks. The Docker build is green now. One concern is that removing --locked means Docker no longer verifies uv.lock, so image dependency resolution can drift between builds. I verified locally that adding uv's relative cutoff metadata (exclude-newer-span = P7D) allows the original uv sync --locked command and the full Docker build to pass. Was dropping lock enforcement intentional, or would you prefer a follow-up PR that restores --locked with the lock metadata synchronized?

PR #2439, opened 2026-07-31
feat: record subdomain collection outcomes

NotoriousRebel:
## Summary

- route every declared `SUBDOMAINS` source through one collection seam
- normalize and validate IDNA hostnames, retain neutral in-scope and out-of-scope observations, and preserve legacy `host` storage
- record completed adapters as succeeded or empty, and record propagated adapter exceptions as failed
- make crt.sh, BuiltWith, and SecurityScorecard propagate transport, non-2xx, redirect, and parsing failures instead of presenting them as valid empty results
- preserve non-host routes when only subdomain collection is unusable

## Operator impact

Malformed hostnames are not silently stored. Failures surfaced by an adapter remain visible, including crt.sh, BuiltWith, and SecurityScorecard provider failures. Existing CLI, REST, file, and stash consumers continue receiving legacy in-scope subdomain values. No DNS resolution is enabled by this change.

## Adapter boundary

The collection seam cannot infer failures that a legacy adapter catches internally. Updating every provider adapter is intentionally outside this focused layer; those adapters must propagate their own failures before the seam can label them failed.

## Scope

This is a source collection layer, not a completed enumeration run. It intentionally excludes run IDs, timestamps, source families, corroboration, merged entities, new persistence, and output formats.

Based directly on merged #2451 and supersedes the original #2439 implementation.

## Verification

- focused tests: 31 passed
- full suite: 262 passed, 6 skipped
- Ruff, formatting, mypy, and git diff checks passed
- zero em dashes
- parallel Standards and Spec reviews passed against `d5628110` at `2d89437`

### DIFF
--- a/tests/discovery/test_builtwith.py
+++ b/tests/discovery/test_builtwith.py
@@ -79,6 +79,7 @@ async def test_process_accepts_text_json_content_type(monkeypatch) -> None:
     await search.process()
 
     assert await search.get_hostnames() == {'sub.example.com'}
+    assert await search.get_interestingurls() == {'https://example.com/login'}
     assert await search.get_interesting_urls() == {'https://example.com/login'}
     assert await search.get_frameworks() == {'Django'}
     assert await search.get_languages() == {'Python'}
@@ -88,7 +89,7 @@ async def test_process_accepts_text_json_content_type(monkeypatch) -> None:
 
 
 @pytest.mark.asyncio
-async def test_process_handles_non_200_status(monkeypatch) -> None:
+async def test_process_reports_non_200_status(monkeypatch) -> None:
     monkeypatch.setattr(builtwith.Core, 'builtwith_key', lambda: 'dummy-key')
 
     class _FakeResponse:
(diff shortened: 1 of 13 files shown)


L1ghtn1ng:
Compatibility issue: organization targets can now abort every result route on a multi-route source.

`execute_collection()` validates `target` before entering its source-execution `try` block (`theHarvester/lib/run.py:62-68`). For a supported invocation such as `-d "Acme Inc" -b baidu`, `normalize_hostname()` returns `None`, the `ValueError` escapes through `store()`, and the worker drops the entire coroutine. As a result, `get_emails()` is never called even though Baidu declares both `SUBDOMAINS` and `EMAILS`. The existing CLI contract explicitly accepts a company name or domain, and previously non-host routes still ran even when no scoped hostname could be produced.

Please keep source processing and non-host getters available when the target is not a DNS hostname, and limit the validation failure to the subdomain evidence route. A regression test with an organization-name target and a source that declares `SUBDOMAINS` plus `EMAILS` would cover the compatibility boundary.

NotoriousRebel:
Closing this draft in its current form to reduce review debt. The reviewed source-collection work is retained in the fork tracker and branches; any future upstream submission will be reconstructed as a smaller independent PR after the organization-target compatibility gap is resolved. The `feature/evidence-runs` branch is retained.

PR #2441, opened 2026-07-31
fix: preserve incomplete DNSDB evidence

NotoriousRebel:
## Summary

- retain DNSDB findings returned before a provider failure
- classify incomplete source execution explicitly
- adapt the existing DNSDB public seam without requiring a live API key in routine tests

## Operator impact

Operators keep useful partial findings and can see that the provider run was incomplete instead of mistaking partial evidence for a complete query.

## Stack

Layer 3 of 10. Base: feature/evidence-source-families.

## Verification

Routine tests use fake HTTP responses. DNSDB live testing remains unavailable without an API key. The complete stack passed all required offline checks.

### DIFF
--- a/tests/discovery/test_dnsdb.py
+++ b/tests/discovery/test_dnsdb.py
@@ -1,11 +1,19 @@
 from __future__ import annotations
 
+import asyncio
 import logging
 
 import pytest
 
 from theHarvester.discovery import dnsdb
 from theHarvester.discovery.constants import MissingKey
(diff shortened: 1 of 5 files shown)


NotoriousRebel:
Closing this stacked draft to reduce review debt. The useful DNSDB outcome work is retained on `fix/dnsdb-evidence-status` and can be reconsidered later as a bounded independent provider migration.

PR #2445, opened 2026-07-31
feat: integrate completed run reporting

NotoriousRebel:
## Summary

- connect CLI collection and every selected post-source stage to RunResult completion
- adapt REST responses and structured persistence from the completed result
- report only after DNS, takeover, Shodan, screenshot, and API stages finish when selected

## Operator impact

Terminal and file output describe the run that actually finished, including its effective options, instead of reporting before later stages have contributed results.

## Stack

Layer 7 of 10. Base: feature/completed-output-adapters.

## Verification

Integration tests cover stage ordering, REST adaptation, compatibility output, and persistence. The complete stack passed all required checks and both reviews.

### DIFF
--- a/tests/discovery/test_intelxsearch.py
+++ b/tests/discovery/test_intelxsearch.py
@@ -142,7 +142,12 @@ async def test_orchestrator_stores_intelx_hostnames(monkeypatch: pytest.MonkeyPa
         async def get_interestingurls(self) -> list[str]:
             return []
 
+    class _RunStore:
+        async def save(self, _result) -> None:
+            return None
+
     monkeypatch.setattr(theharvester_main.stash, 'StashManager', _Stash)
+    monkeypatch.setattr(theharvester_main, 'SQLiteRunStore', _RunStore)
     monkeypatch.setattr(intelxsearch, 'SearchIntelx', _Intelx)
 
     results = await theharvester_main.start(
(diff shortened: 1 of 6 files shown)


L1ghtn1ng:
Two completed-run integration paths still lose supported behavior/data:

1. `completed_run_result = await execute_run(word, ())` runs for every invocation without a migrated evidence source (`theHarvester/__main__.py:1475-1477`), while `execute_run()` rejects any target that is not a hostname. Therefore a supported organization query such as `-d "Acme Inc" -b rocketreach` can finish its provider work and then fail during finalization, even though RocketReach has no hostname route. Please preserve the CLI's company-name contract in the run model, and only require a DNS hostname for hostname scoping/validation.

2. DNS brute force returns `resolved_pair, hosts, ips`, but the stage records only `HOSTNAME` findings from `resolved_pair` (`theHarvester/__main__.py:1549-1554`). `complete_run()` then splits each value at the first colon and retains only the hostname. The returned `ips` are never represented in the completed result, so the later JSONL stream contains `late.example.com` but no `ip-address` record for `192.0.2.11`. Please add explicit `IP_ADDRESS` findings from `ips` (and retain the host/address association if needed), with coverage through the completed output/persistence adapters.

NotoriousRebel:
Closing this stacked draft to reduce review debt. The integration still has unresolved organization-target and DNS brute-force IP preservation findings. The `feature/completed-run-integration` branch is retained for evidence.

PR #2447, opened 2026-07-31
fix: preserve incomplete provider evidence

NotoriousRebel:
## Summary

- start the evidence run before provider collection
- preserve partial Common Crawl and Cert Spotter findings with incomplete status
- fail closed when providers cannot prove completion
- classify selected operator actions separately from passive collection

## Operator impact

Operators can distinguish empty, complete, partial, and failed provider work while retaining useful findings already returned.

## Stack

Layer 9 of 10. Base: fix/completed-run-finalization.

## Verification

Focused regressions cover Common Crawl, Cert Spotter, action classification, and run start ordering. The complete stack passed all required checks and both reviews.

### DIFF
--- a/tests/discovery/test_certspotter.py
+++ b/tests/discovery/test_certspotter.py
@@ -9,6 +9,7 @@ import pytest
 
 from theHarvester.discovery import certspottersearch
 from theHarvester.lib.core import Core
+from theHarvester.lib.run import SourceIncompleteError, SourcePartialError, SourceRateLimitedError
 
 
 class TestCertspotter(object):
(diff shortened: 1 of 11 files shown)


L1ghtn1ng:
Mixed valid/malformed Common Crawl pages are still reported as complete.

`_safe_parse_json_lines()` sets a local `malformed` flag, logs a warning, and returns the valid records (`theHarvester/discovery/commoncrawl.py:43-63`). The caller has no indication that a record was dropped and sets `query_succeeded = True` after consuming the returned records. I reproduced this with a payload containing one valid URL line plus `not-json`: `process()` returned normally with `api.example.com`, so `execute_run()` would record `SUCCEEDED` rather than `PARTIAL` even though the provider response was incomplete.

Please propagate the partial condition while retaining the valid records (for example, return a parse-status flag or raise a partial error carrying the accumulated findings), and add a regression that asserts the source execution is `PARTIAL` with the valid hostname preserved.

NotoriousRebel:
Closing this stacked draft to reduce review debt. Mixed valid and malformed Common Crawl records still need a truthful partial-outcome contract in a bounded provider migration. The `fix/completed-evidence-semantics` branch is retained.

PR #2453, opened 2026-08-01
Fix screenshot paths on macOS

Rahulk644:
## Summary

- treat only `win*` platform identifiers as Windows when constructing screenshot output paths
- add regression coverage for Darwin, Linux, and Windows platform strings

## Root cause

The screenshot module selected the Windows path separator with `"win" in sys.platform`. On macOS, `sys.platform` is `darwin`, which also contains the substring `win`. As a result, screenshot output paths used backslashes on macOS and could create files with backslashes in their names instead of placing them in the requested directory.

Using `sys.platform.startswith("win")` preserves the Windows behavior without misclassifying Darwin.

## Validation

- `pytest`: 123 passed
- `ruff check`: passed
- `ruff format --check`: passed
- `mypy`: passed for the changed source module
- `git diff --check`: passed

### DIFF
--- a/tests/test_screenshot.py
+++ b/tests/test_screenshot.py
@@ -1,0 +1,21 @@
+from __future__ import annotations
+
+import pytest
+
+from theHarvester.screenshot import screenshot
+
+
+@pytest.mark.parametrize(
+    ('platform', 'expected_separator'),
+    [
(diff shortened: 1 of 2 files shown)


NotoriousRebel:
Good catch, thanks!

PR #2487, opened 2026-08-06
Add catalog-driven passive collection and sources

NotoriousRebel:
## Summary

This is one review batch containing three focused commits:

1. Add an explicit authenticated HIBP verified-domain CLI source while preserving the keyless public HIBP source. Broad selectors cannot invoke the authenticated endpoint.
2. Make the source catalog drive CLI help, P0-only `all` selection, and the P0/P1/P2 activity summary. Remove duplicate late provider execution.
3. Add a bounded keyless Arquivo.pt passive source with defensive JSONL parsing and scoped hostname normalization.

The commits are also available as a dependent fork stack for layer-by-layer inspection:

- NotoriousRebel/theHarvester#249
- NotoriousRebel/theHarvester#250
- NotoriousRebel/theHarvester#251

## Verification

- 424 passed, 6 skipped
- `ruff check .`
- `ruff format --check .`
- `mypy theHarvester`
- `git diff --check`
- zero em dashes in the aggregate diff
- parallel Standards and Spec reviews passed against `ffd03b13`

An authorized live Arquivo.pt CLI smoke against `mozilla.org` returned `www.mozilla.org`; the result was confirmed in JSON, JSONL, and SQLite completed results. HIBP verified-domain testing remained offline because live use requires a paid user-owned key and a domain verified in that HIBP account.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Allowed REST `/query` requests to select discovery sources by result capability, matching the CLI's union semantics while preserving explicit source selection.
+- Changed `-b all` to select every cataloged P0 passive source once while leaving P1 DNS and P2 direct sources available through explicit selection.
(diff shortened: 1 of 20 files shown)


NotoriousRebel:
Superseded by the requested three-layer upstream stack. Replacement PR links will be added after gh stack submission.

PR #2493, opened 2026-08-06
Add truthful DNS source contracts and bounded recursion

NotoriousRebel:
## Summary

- harden the Pentest-Tools, Shodan InternetDB, Subdomain Finder C99, and Windvane DNS source contracts
- add three-vantage DNS consensus with closest-encloser wildcard controls
- add bounded breadth-first recursive DNS discovery to the CLI and REST API
- retain meaningful DNS classifications in completed JSONL and SQLite results while keeping legacy hostname, IP, JSON, and XML contracts current-only

## Why

Several direct DNS-backed sources mixed malformed provider values, terminal failures, and valid hostnames. Recursive DNS also needs an explicit safety and truth boundary so guessed labels, wildcard answers, and resolver disagreement are not promoted as current hosts.

This change normalizes each source at its adapter boundary and adds recursive discovery that advances only candidates with two-resolver address consensus. It stops on configured depth, query, and runtime limits or three consecutive zero-yield batches.

## Safety

- requires exactly three canonically distinct resolver IPs
- follows CNAMEs only inside the authorized target scope
- records external CNAMEs as context without querying them
- omits consensus-negative guessed labels while retaining current, wildcard, disputed, dangling-alias, NODATA, and other DNS-existing classifications
- performs no live reconnaissance during automated tests

## Validation

- `pytest`: 476 passed, 6 skipped
- focused DNS, persistence, CLI, and REST tests: 33 passed
- Ruff check and formatting check
- mypy
- `git diff --check`
- parallel Standards and Spec reviews with no remaining findings

A bounded authorized live validation against `mozilla.org` used three public resolver vantages. The recursive engine completed depth 1 with 288 accounted queries and classified `www.support.mozilla.org` as currently addressable while retaining its scoped CNAME evidence.


NotoriousRebel:
Superseded by focused PRs #2494 (direct provider contracts), #2495 (DNS consensus), and #2496 (RapidDNS resolution parity). The dependent recursive engine, CLI/persistence, and REST layers remain split on origin and will be promoted one at a time after their prerequisite merges.

PR #2508, opened 2026-08-08
Add an authenticated durable run API v1

NotoriousRebel:
## Summary

- replace the unversioned and provider-specific REST surface with one authenticated `/api/v1` source and run contract, without compatibility routes or redirects
- persist queued, running, completed, failed, and cancelled runs in SQLite while retaining partial evidence across cancellation, deadlines, and restart recovery
- share source selection and enumeration semantics with the CLI while preserving normal DNS, redirects, private targets, and public targets
- make versioned JSONL the API's only file import and export format; preserve evidence identity, UTC timestamps, status, source outcomes, DNS status, and canonical `hostname` and `ip-address` wire kinds across round trips
- reject malformed JSONL UUIDs, timestamps, chronology, and finding shapes before persistence; refuse export until evidence exists
- derive screenshot paths only from persisted run evidence, with symlink and traversal regressions

ADR 0003 remains `proposed` until this policy and implementation are accepted upstream.

## Stack

Layer 1 of 3. Merge this PR first. The HarvestView UI and container release follow on descendant branches.

## Verification

- 580 pytest tests passed, 6 skipped at the complete publication head
- focused API and lifecycle regressions, including strict JSONL round trips and sparse partial-evidence export
- Ruff and formatting check
- mypy
- `git diff --check`
- zero em dashes
- parallel Standards and Spec reviews with no findings

The complete stack was also exercised through a bounded live Cert Spotter run against `mozilla.org`; the run completed with 241 scoped results and persisted across restart.


NotoriousRebel:
Superseded by #2511, which carries the same reviewed API head as layer 1 of native GitHub stack #2514. Closing this fork-headed PR so review continues on the correctly chained stack. Nothing was merged here.

PR #2509, opened 2026-08-08
Add the HarvestView web application

NotoriousRebel:
## Summary

- serve HarvestView at `/` as the browser workspace for creating, monitoring, cancelling, importing, and exporting API v1 runs
- keep the operator API key out of browser storage by issuing a derived HttpOnly local session cookie
- provide source and capability selection, run history, normalized results, source execution state, screenshots, and honest cancellation or failure states
- make `.jsonl` the only file picker format and provide one `All JSONL` export, including terminal runs with zero findings; remove JSON and CSV export controls and client code
- remove the former marketing root and leave `/app` intentionally unavailable
- add responsive and accessibility behavior plus real Chromium regression coverage
- remove unused Bootstrap and its Bootstrap-specific Tabulator theme
- remove the redundant `THIRD_PARTY.md` notice while retaining the required `TABULATOR-LICENSE`

## CSS decision

HarvestView keeps its purpose-built `app.css`; adopting Bootstrap, Bulma, Pico, or Tailwind would duplicate that design layer or require a markup and build-pipeline rewrite. Tailwind's browser CDN is also officially development-only.

Tabulator 6.5.2's JavaScript and default table theme remain same-origin, so HarvestView and its browser tests do not depend on a CDN. The README records the exact checksum plus an optional connected-deployment CDN URL, SRI hash, privacy tradeoff, and required attributes.

## Stack

Layer 2 of 3. Depends on #2508 and should merge after it. This branch descends from `codex/api-v1-run-lifecycle`; after #2508 merges, this PR shrinks to the HarvestView layer.

## Verification

- two consecutive complete runs of all 17 Chromium end-to-end tests
- desktop and mobile viewports, light and dark themes, no external browser requests, and restart-safe test lifecycle
- 580 pytest tests passed, 6 skipped at the complete publication head
- exact SHA-256 verification of the pinned same-origin Tabulator stylesheet
- Ruff and formatting check
- mypy
- `git diff --check`
- zero em dashes
- parallel Standards and Spec reviews with no findings

The complete stack was also exercised through a bounded API v1 Cert Spotter run against `mozilla.org` with complete, durable evidence.

PR #2510, opened 2026-08-08
Package HarvestView for local deployment

NotoriousRebel:
## Summary

- build a pinned multi-stage image that runs the installed wheel as UID 10001 and includes Chromium screenshot support
- force uv to rebuild the copied local `theharvester` package so a persistent Docker cache cannot install stale project code when the version is unchanged
- publish HarvestView only on host loopback through Compose, with authenticated API health, a read-only root filesystem, and durable private run storage
- load the operator key from a read-only file secret while excluding provider keys, proxies, local databases, and runtime state from image layers
- add a dedicated container smoke workflow covering routes, authentication, Chromium launch, JSONL import and export, restart, and persisted evidence
- update README, wiki, and changelog for the final `/api/v1` and HarvestView deployment contract

The container workflow is separate from the browser E2E workflow, so this layer does not rewrite the UI layer's CI file.

## Stack

Layer 3 of 3. Depends on #2509, which depends on #2508. Merge this PR last. This branch descends from both earlier layers and its diff shrinks as they merge.

## Verification

- fresh Docker image build with an explicit local-package rebuild in the uv log
- installed container source matched the publication-head JSONL exporter
- localhost smoke: `/` returned 200, `/app` returned 404, and authenticated `/api/v1/runs` returned 200
- real container JSONL import and export preserved the evidence UUID and emitted canonical `hostname` and `ip-address` records
- healthy authenticated API probe, UID 10001, installed-wheel provenance, Chromium 148 headless launch, and restart persistence
- 580 pytest tests passed, 6 skipped, plus two consecutive complete 17-test Chromium runs
- Ruff, formatting, mypy, Compose config, workflow YAML, `git diff --check`, and zero em dashes
- parallel Standards and Spec reviews with no findings

This PR does not change discovery-provider behavior or impose a public-only target policy.


NotoriousRebel:
Superseded by #2513, which carries the same reviewed container head as layer 3 of native GitHub stack #2514. Closing this fork-headed PR so review continues on the correctly chained stack. Nothing was merged here.

PR #2515, opened 2026-08-08
Use SQLAlchemy for async result persistence

NotoriousRebel:
## Summary

This PR replaces direct SQLite persistence calls with one `ResultStore` module built on SQLAlchemy 2 ORM models and `AsyncSession` over the `aiosqlite` driver.

It also:

- deletes the old `stash.py` facade and moves its behavior into `database.py`;
- gives CLI and REST callers one persistence interface: initialize, save, load, list, and transitional runless observation writes;
- keeps SQLAlchemy models, sessions, and exceptions private to the persistence implementation;
- keeps one path-owned async engine and session factory;
- records schema version 1 and rejects databases created by a newer application;
- upgrades released databases transactionally without inventing run IDs or dropping incomplete historical rows;
- stores canonical result kinds and source names;
- keeps SQLite and JSONL terminal results in parity;
- enables WAL only on SQLite builds containing the upstream WAL-reset fix;
- verifies foreign-key enforcement on the actual runtime engine and fails initialization if it is disabled;
- preserves cascades, ordering, atomic writes, and dotted-parent output paths.

SQLAlchemy still uses `aiosqlite` as its async SQLite driver. Application code no longer imports or calls `aiosqlite` directly.

## Persistence interface

Callers use `ResultStore` from `theHarvester.lib.database`:

```python
store = ResultStore()
await store.initialize()
await store.save_run(completed_result)
recent = await store.list_runs(limit=50)
stored = await store.load_run(completed_result.run_id)
```

One enumeration reuses one store instead of constructing separate managers for source, DNS, API-scan, and terminal writes. SQLAlchemy row types and sessions stay inside `database.py`. Duplicate IDs raise `DuplicateRunError`; other SQLAlchemy storage failures cross the interface as `ResultStoreError` with the original exception chained.

The existing filename remains `~/.local/share/theHarvester/stash.sqlite` so upgrades find user data in place. The obsolete Python module and `StashManager` class are removed with no compatibility alias.

## Database model

This foundation contains:

- `runs`: one completed enumeration run and its timestamps;
- `results`: the deduplicated terminal results for that run, in output order;
- `discovery_observations`: source observations retained from the released runless persistence behavior.

Normal CLI and API initialization upgrades an existing database in one transaction. It temporarily renames the released `results` table, moves `completed_results` and `completed_result_items` to `runs` and `results`, copies historical source rows into `discovery_observations`, removes the temporary table, and sets schema version 1.

The released observation table allowed null values. The migration preserves those incomplete rows losslessly instead of failing initialization, deleting them, or inventing provenance. New observation writes still provide normalized non-null values.

Users do not need to run a migration command.

## WAL behavior

WAL lets readers continue while a writer commits and generally reduces write synchronization. SQLite still permits only one writer at a time.

`uv.lock` pins Python packages but cannot pin the SQLite C library embedded in each Python runtime. This PR enables WAL only on fixed SQLite versions: 3.44.6+, 3.50.7+, or 3.51.3+. Other supported versions use DELETE journal mode.

## Verification

- Focused persistence tests: 20 passed
- Full pytest: 606 passed, 6 skipped
- Ruff check: passed
- Ruff formatting check: passed
- mypy: passed
- uv lock check: passed, 89 packages
- git diff check: passed
- Zero em dashes in changed files
- Parallel Standards review: no findings
- Parallel Spec review: no findings

An earlier bounded crt.sh smoke test against `mozilla.org` produced 241 hostname results and confirmed the end-to-end collection, JSONL, and SQLite path. The final amendments are persistence-interface and migration changes, so their verification is offline and migration-focused.

## Scope

This is the SQLAlchemy, WAL, migration, and persistence-interface foundation. It does not add the proposed `executions`, `result_origins`, `artifacts`, relationships, JSONL v2, REST redesign, or HarvestView work. Those require truthful runtime producers and remain dependent reviewable changes.


NotoriousRebel:
## Updated reviewer walkthrough

The persistence implementation now has one module and one caller-facing class.

### Start here

1. `theHarvester/lib/database.py` contains `ResultStore`, private ORM rows, engine lifecycle, the released-schema migration, WAL selection, and runtime foreign-key validation.
2. `theHarvester/__main__.py` reuses one `ResultStore` throughout an enumeration.
3. `theHarvester/lib/api/api.py` uses the same interface for completed-run routes and lifecycle initialization.
4. `tests/lib/test_completed_persistence.py` covers released upgrades, null historical rows, rollback, restart, concurrency, exception translation, ordering, WAL, foreign keys, cascades, JSONL parity, and event-loop behavior.

`theHarvester/lib/stash.py` and `StashManager` are gone. There is no compatibility alias and callers do not import SQLAlchemy models or sessions.

### Existing databases

The SQLite filename remains `~/.local/share/theHarvester/stash.sqlite` intentionally. Changing that path would strand existing data. Initialization finds the released tables there and migrates them transactionally.

Historical source rows do not have run IDs, so the migration does not attach them to fabricated runs. The released columns were nullable, and the migration now preserves incomplete historical rows as well as complete ones.

### What this PR does not claim

This is not yet the five-concept provenance model. `executions`, `result_origins`, and `artifacts` need the collection runtime to emit truthful source/action outcomes and artifact metadata. Adding empty tables here would produce a misleading schema. Those changes remain in the dependent provenance stack, before the REST and HarvestView work is rebuilt on top.

Exact reviewed head: `131ed855695f71091b2039c53c57a15618b3fa1f`.

Final local verification: 606 passed, 6 skipped; Ruff, formatting, mypy, uv lock, diff check, and both parallel review axes are clean.

PR #2519, opened 2026-08-11
fix: extract hostnames from bufferover.run responses correctly

SkxOverKill:
## Summary

\SearchBufferover.do_search()\ crashes with an \IndexError\ on every real response from \	ls.bufferover.run\, so the bufferoverun source never returns any hosts.

The hostname extraction in \do_search\ was broken in three ways:

1. **Crash:** entries were indexed at \host.split(',')[4]\, but the API returns four comma-separated fields per line (\<ip>,<sha256>,<cert-org>,<CN/SNI>\), so index 4 is always out of range.
2. **Chained comparison bug:** the guard used \self.word in host.split(',')[0] in host\, which Python evaluates as \A in B and B in C\. The extra \host.split(',')[0] in host\ clause is always true (a field is always a substring of its own line) and the word was checked against the IP field rather than the hostname.
3. **Type mix:** the "true" branch of the old ternary returned the full \host.split(',')\ list into the \	otalhosts\ set, mixing lists and strings.

The IP extraction (\	otalips\) was also run unconditionally, so a failed/empty fetch (an empty string or a dict without \Results\) raised a \TypeError\ instead of returning gracefully.

## Changes

- Extract the hostname from the last field of each line (\
split(',', 1)\) so commas in the certificate-organization field cannot shift the index, matching the documented API format.
- Return early when the response is empty or lacks \Results\, so failed fetches no longer raise.
- Keep the existing IP extraction but gate it behind the same guard.

## Test

Added \	ests/discovery/test_bufferoverun.py\ with six cases covering hostname/IP extraction, commas in the cert-org field, and empty/failed responses. All 128 api tests pass (1 pre-existing network-dependent failure in \	est_thc\), ruff check and format are clean.

### DIFF
--- a/tests/discovery/test_bufferoverun.py
+++ b/tests/discovery/test_bufferoverun.py
@@ -1,0 +1,80 @@
+import pytest
+
+from theHarvester.discovery import bufferoverun
+
+
+def _patch_fetch(monkeypatch, payload):
+    import theHarvester.lib.core as core_module
+
+    async def fake_fetch_all(urls, headers=None, params='', json=False, takeover=False, proxy=False):
+        return [payload]
+
+    monkeypatch.setattr(core_module.AsyncFetcher, 'fetch_all', staticmethod(fake_fetch_all), raising=True)
+
+
+class TestSearchBufferover:
+    def test_init_sets_state(self, monkeypatch):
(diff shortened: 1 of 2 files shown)


NotoriousRebel:
Can you test if this change is necessary against dev branch as I believe this issue has been addressed.

NotoriousRebel:
Closing as this has been fixed in dev branch.

PR #2529, opened 2026-08-12
baidusearch: remove forced @ (%40) prefix from Baidu query

SkxOverKill:
## Summary

Contributes to #2403 (Baidu module broken). The module hardcodes `%40` (the `@` symbol) into the search query (`wd=%40{word}`), which is not valid Baidu syntax and turns every query into a different search than intended. Fixes the query string to issue a plain `wd={word}` search.

This is the minimal "query string fix" from the issue; it complements the playwright-based rework the maintainer is considering, and is a strict improvement on its own.

## Changes

- `theHarvester/discovery/baidusearch.py`: drop the forced `%40` prefix from `wd=`.
- `tests/discovery/test_baidusearch.py`: update expected URLs to match the fixed query string.

## Verification

- `pytest tests/discovery/test_baidusearch.py` -> 2 passed
- `ruff check theHarvester/discovery/baidusearch.py` -> All checks passed

### DIFF
--- a/tests/discovery/test_baidusearch.py
+++ b/tests/discovery/test_baidusearch.py
@@ -29,9 +29,9 @@ class TestBaiduSearch:
         await search.process(proxy=True)
 
         expected_urls = [
-            "https://www.baidu.com/s?wd=%40example.com&pn=0&oq=example.com",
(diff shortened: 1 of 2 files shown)


NotoriousRebel:
This has been fixed in dev branch, closing.

PR #2540, opened 2026-08-14
refactor: remove implicit Windvane DNS guessing

NotoriousRebel:
## Summary
- remove the Windvane adapter and all catalog, dispatcher, credential, and configuration registration
- eliminate its unauthenticated fallback that implicitly guessed and resolved common DNS names
- keep DNS brute force available only through the explicit operator action
- reject the retired source with clear CLI and REST migration guidance

## Verification
- upstream-base affected suite: 244 passed
- upstream-base full offline suite: 1202 passed, 6 skipped, 26 deselected
- Ruff, format, mypy, and diff checks passed
- independent Standards and Spec reviews: clear

No live target traffic was used.


NotoriousRebel:
Closing because the scope was misinterpreted. Windvane should remain; only its unauthenticated fallback that guesses common DNS labels should be removed. A corrected focused PR will follow.

PR #2575, opened 2026-08-18
Fix Playwright startup on Windows

NotoriousRebel:
## Root cause

`winloop` 0.6.3 rejects the `startupinfo` argument Playwright passes to `asyncio.create_subprocess_exec` when launching its driver. This affects both Baidu's Playwright path and `--screenshot` browser startup.

## Fix

- retain `winloop` for ordinary Windows runs
- select standard asyncio before startup when `--screenshot` or `--screenshot=...` is present
- fall back to Baidu's existing aiohttp path only for the exact `ValueError: startupinfo is not supported` failure
- keep `uvloop` behavior unchanged on non-Windows platforms
- restore the pinned Windows `winloop` dependency and regression coverage

## Native Windows benchmark

Windows 11, Python 3.14.4, winloop 0.6.3; 6 alternating process pairs and 18 measured trials per loop:

| Measurement | Standard asyncio | winloop | Difference |
|---|---:|---:|---:|
| Scheduler elapsed | 127.10 ms | 69.69 ms | 45.2% less time / 82.2% higher throughput |
| Local HTTP throughput | 8,887 req/s | 13,906 req/s | 57.4% higher |
| HTTP p50 latency | 10.21 ms | 6.71 ms | 34.3% lower |
| HTTP p95 latency | 11.12 ms | 7.60 ms | 31.6% lower |
| Playwright-style subprocess | 6/6 passed | 0/6 passed | winloop raises `startupinfo is not supported` |

The loopback workload exaggerates the benefit compared with remote provider traffic, but the repeated paired result is large enough to justify retaining winloop outside Playwright-dependent paths.

The actual `ScreenShotter._browser()` path was also tested natively: it fails under winloop 0.6.3 and succeeds under standard asyncio.

## Verification

- native Windows bootstrap tests: 7 passed
- focused Baidu and screenshot tests: 39 passed
- full non-browser suite: 1,790 passed, 6 skipped, 33 deselected
- `uv lock --check`
- `uv run ruff check .`
- `uv run mypy theHarvester` (WSL): success, 111 source files

Native Windows mypy still reaches the existing `run_worker.py` use of `os.killpg`/`signal.SIGKILL`; that platform-specific typing issue is unrelated to this change.

## Remaining risk

Baidu's aiohttp fallback may be less resilient to anti-bot pages than Playwright. The fallback is restricted to the exact known winloop startup incompatibility; unrelated `ValueError`s still propagate.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ dependencies = [
     "sqlalchemy==2.0.51",
     "uvicorn==0.52.1",
     "uvloop==0.22.1; platform_system != 'Windows'",
-    "winloop==0.6.3; platform_system == 'Windows'",
 ]
 
 [dependency-groups]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ dependencies = [
     "sqlalchemy==2.0.51",
     "uvicorn==0.52.1",
     "uvloop==0.22.1; platform_system != 'Windows'",
+    "winloop==0.6.3; platform_system == 'Windows'",
 ]
 
 [dependency-groups]
--- a/tests/discovery/test_baidusearch.py
+++ b/tests/discovery/test_baidusearch.py
@@ -287,6 +287,23 @@ class TestBaiduSearch:
         assert http.closed
 
     @pytest.mark.asyncio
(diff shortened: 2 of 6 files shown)


NotoriousRebel:
Holding off from merging as plan to do more benchmarking on Windows to see what benefits winloop offers in terms of speed and if there is an alternative route to preserve it.

PR #2586, opened 2026-08-25
fix: enforce discovery security boundaries

L1ghtn1ng:
## Summary

- keep one proxy identity across each provider conversation and prevent silent proxy bypasses
- route IntelX, THC, and RocketReach through the shared proxy-aware transport
- preserve the complete explicit hostname target boundary, including a leading `www` label
- create credential configuration atomically with owner-only permissions
- canonicalize hostname and IP evidence before persistence or JSONL export
- retain case-insensitive compatibility for legacy CLI source identifiers

## Verification

- focused core, evidence, and runner tests: 287 passed
- focused provider, scope, and CLI regressions: 55 passed, 4 skipped
- full host-side suite: 1962 passed, 6 skipped; three unrelated workflow-contract tests failed because the implementation checkout used the current `ty` configuration while those tests expected `mypy`
- affected Shodan regression corrected and focused rerun: 2 passed
- `uv run ruff check .`
- `uv run ty check`
- `git diff --check`


NotoriousRebel:
I rechecked this branch against the current `dev` and still see three issues. The two proxy behaviors already exist on `dev`, so they are not regressions caused by this branch. They do leave the PR's proxy claim incomplete.

1. Explicit proxy mode still fails open.

   [`_resolve_proxy(True)`](https://github.com/laramies/theHarvester/blob/f2efcb5b99f6d83f6d6a38a101730ca2189d30b6/theHarvester/lib/core.py#L497-L505) returns `(None, None)` when the configured proxy list has no usable entries. [`open_session`](https://github.com/laramies/theHarvester/blob/f2efcb5b99f6d83f6d6a38a101730ca2189d30b6/theHarvester/lib/core.py#L533-L550) then creates a normal direct session. A run started with `--proxies` can therefore send provider traffic directly without reporting an error or a failed or partial source outcome.

   The transport should fail closed when proxy use is explicitly requested and return a sanitized `proxy-unavailable` outcome. A regression test with an empty proxy list would cover this case.

2. The proxy is stable for one helper call, not the whole provider conversation.

   [`fetch_all`](https://github.com/laramies/theHarvester/blob/f2efcb5b99f6d83f6d6a38a101730ca2189d30b6/theHarvester/lib/core.py#L968-L1011) and [`post_fetch`](https://github.com/laramies/theHarvester/blob/f2efcb5b99f6d83f6d6a38a101730ca2189d30b6/theHarvester/lib/core.py#L685-L727) can each create a new session and select another proxy. That lets dependent requests switch proxies during a single provider operation. CriminalIP, for example, starts a scan, polls its status, and fetches the final report through separate helper calls ([scan, poll, and report flow](https://github.com/laramies/theHarvester/blob/f2efcb5b99f6d83f6d6a38a101730ca2189d30b6/theHarvester/discovery/criminalip.py#L89-L176)).

   The adapter should own one session for the complete operation and pass it to every request. A regression test with two configured proxies should confirm that the flow makes only one selection.

3. The hostname scope decision is not recorded.

   [`CONTEXT.md`](https://github.com/laramies/theHarvester/blob/f2efcb5b99f6d83f6d6a38a101730ca2189d30b6/CONTEXT.md#L9-L13) requires changes to a release boundary to update both the context and the nearest decision record in the same PR. This branch adds the exact leading `www.` authorization rule to `CONTEXT.md` and changes the implementation, but it does not add or update a decision record under `docs/adr/` or an equivalent location.

   This needs a short decision record explaining why the operator-supplied DNS name is the exact authorization boundary, why `www.` is not an alias for the apex domain, and why any provider-specific apex lookup must stay separate from result scope enforcement.

The second commit gets the active `www.` behavior right and centralizes hostname and IDNA validation. The unused legacy `theHarvester/parsers/securitytrailsparser.py` still strips `www.`, but the current SecurityTrails adapter does not import it. That dead-code cleanup can stay separate.

PR #2593, opened 2026-08-27
Refactor API background state into app lifespan

NotoriousRebel:
## Summary

- replace module-level worker and scheduler lifecycle globals with app-owned `RunWorker`, `SchedulerService`, and `ApiRuntime` instances
- inject the per-lifespan runtime through `create_app(runtime_factory=...)` and FastAPI dependencies while preserving the durable SQLite worker lease and public contracts
- make child cancellation cleanup and unexpected background-task failure reporting explicit, with focused isolation and lifecycle coverage
- record the ownership decision in ADR 0010 (0010 avoids colliding with the proxy-scope ADR already present on `master`)

## Verification

- `uv sync --all-groups`
- focused runtime/schedule/run/API suites: `169 passed`
- `uv run ruff check .`
- `uv run ty check`
- full non-browser suite: `1988 passed, 6 skipped, 36 deselected`
- HarvestView browser E2E: `36 passed`
- focused documentation checks: `16 passed`

No REST schema, database schema, environment variable, CLI flag, or discovery-activity boundary changes.

### DIFF
--- a/theHarvester/lib/api/run_worker.py
+++ b/theHarvester/lib/api/run_worker.py
@@ -1,82 +1,70 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import os
 import signal
 import subprocess
 import sys
 from argparse import ArgumentParser
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import Any, Protocol
 from uuid import UUID, uuid4
 from weakref import WeakKeyDictionary
-
-import anyio
-
-from theHarvester.lib.enumeration import (
-    DEFAULT_RESULT_START,
-    DEFAULT_SOURCE_WORKERS,
-    EnumerationOptions,
-)
-from theHarvester.lib.resolver_selection import DEFAULT_DNS_RESOLVERS
-from theHarvester.lib.virtual_host import (
-    DEFAULT_VHOST_CONCURRENCY,
-    DEFAULT_VHOST_REQUEST_LIMIT,
-    DEFAULT_VHOST_RUNTIME_SECONDS,
-    DEFAULT_VHOST_TIMEOUT_SECONDS,
-)
 
 from .run_artifacts import ensure_private_directory, read_child_evidence, write_child_evidence
 from .run_store import RunStore
 
-if TYPE_CHECKING:
-    from collections.abc import Awaitable, Callable
-
-_worker_task: asyncio.Task[None] | None = None
-_worker_stop: asyncio.Event | None = None
-_worker_wakeup: asyncio.Event | None = None
-_worker_owner: str | None = None
-_process_groups: WeakKeyDictionary[asyncio.subprocess.Process, int] = WeakKeyDictionary()
+ProcessFactory = Callable[[str, Path, Path], Awaitable[asyncio.subprocess.Process]]
+RunStoreFactory = Callable[[], RunStore]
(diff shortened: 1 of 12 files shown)


theHarvester/lib/api/run_worker.py
```diff
@@ -87,199 +75,330 @@ async def read(stream: asyncio.StreamReader | None) -> bytes:
...
-                str(process.pid),
-                '/T',
-                '/F',
-                stdout=asyncio.subprocess.DEVNULL,
-                stderr=asyncio.subprocess.DEVNULL,
+            session.stop_event.set()
+            session.wakeup_event.set()
+            if task is not None:
+                try:
+                    await asyncio.shield(task)
```
L1ghtn1ng, line 140: **Low severity — restart-after-crash asymmetry.**

After the supervisor task crashes, `_session` is still set and `available` is False. A subsequent `start()` calls `await self.stop()`, and this `await asyncio.shield(task)` re-raises the failed task's exception. Empirically:

```
restart raised: boom        # first start() after crash
second restart: OK          # only the second attempt works
```

The old module-level `start_worker()` restarted cleanly after a crash (it only returned early if the task was still running). This path is unreachable in the current one-start-per-lifespan flow, so impact is limited to future restart/recovery callers and direct embedders — but the first-call-raises / second-call-works asymmetry is surprising. Either catch-and-ignore the completed task's exception in `stop()` when the task already finished, or document that `stop()` (and therefore a recovery `start()`) surfaces the crash. Same pattern exists in `schedule_service.py`.

theHarvester/lib/api/run_worker.py
```diff
@@ -87,199 +75,330 @@ async def read(stream: asyncio.StreamReader | None) -> bytes:
...
+            else:
+                await store.fail(
+                    run_id,
+                    f'Child process exited with status {process.returncode} without terminal completion',
+                    log,
+                    evidence=evidence,
+                )
+        except asyncio.CancelledError:
+            await self._abort_process_execution(process, wait_task, output_task)
+            raise
```
L1ghtn1ng, line 342: **Low severity — dead `except asyncio.CancelledError` clause.**

This handler is byte-identical to the `except BaseException` handler that follows it. Since `CancelledError` is a `BaseException` subclass, the first clause is fully subsumed — a single `except BaseException` suffices.

theHarvester/lib/api/runtime.py
```diff
@@ -0,0 +1,50 @@
...
+
+def create_runtime() -> ApiRuntime:
+    worker = RunWorker()
+    return ApiRuntime(worker=worker, scheduler=SchedulerService(worker))
+
+
+def get_runtime(request: Request) -> ApiRuntime:
+    runtime = getattr(request.app.state, 'runtime', None)
+    if not isinstance(runtime, ApiRuntime):
+        raise RuntimeError('theHarvester API runtime is not initialized')
```
L1ghtn1ng, line 41: **Behavior change note — 500 instead of 503 without lifespan.**

This bare `RuntimeError` surfaces as an unhandled 500 on `POST /api/v1/runs` and the schedule routes when the app is served without its lifespan. Previously, `create_run` without a started worker returned 503 "worker is unavailable". This only affects unsupported usage (serving the app without running the lifespan), and failing closed is defensible — the updated `test_sensitive_endpoints_require_validation` shows this was considered. Flagging so it's a conscious contract choice; an `HTTPException(503)` here would preserve the old failure mode if any embedder depends on it.

PR #2613, opened 2026-09-14
feat(discovery): add Serply search source

googio:
## Summary

theHarvester has search-engine sources for Baidu, Brave, DuckDuckGo, Mojeek and Yahoo, but none that reads Google's index. This adds `serply`, a keyed commercial API that returns Google results, registered exactly the way `bravesearch` registers Brave: one catalog entry, one factory entry, one credential field, and one offline provider contract.

Disclosure: I work on [Serply](https://serply.io/). This is a plain discovery module. It touches no CLI orchestration, persistence or output code, and it is skipped like any other keyed source when no key is configured.

## Integration points

| File | Change |
| --- | --- |
| `theHarvester/discovery/serplysearch.py` | new `SearchSerply`, mirroring `bravesearch.py` |
| `theHarvester/lib/source_catalog.py` | `_spec('serply', SUBDOMAINS, EMAILS)` |
| `theHarvester/lib/source_runner.py` | import plus one `SOURCE_FACTORIES` entry |
| `theHarvester/lib/core.py` | `'serply': ('key',)` and a `serply_key()` accessor |
| `theHarvester/data/api-keys.yaml` | key stanza |
| `README.md`, `docs/wiki/Configuration-and-API-Keys.md`, `CHANGELOG.md` | source matrix row, counter, provider note, changelog line |
| `tests/discovery/test_serply.py` | provider contract |
| `tests/test_readme.py`, `tests/lib/test_core.py` | catalog-derived roster assertions that every new source has to update |

## Provider behaviour worth knowing

Three things about this API are easy to get wrong, so each one has a test:

1. **`start` is the only pagination parameter Serply honours.** `page`, `offset` and `p` are accepted and silently ignored. I confirmed this against the live API: `start=10` returns a window with zero overlap, while `page=2` and `offset=10` each return page one verbatim. A loop built on any of those re-fetches the first window forever and spends one credit per duplicate page.
2. **A query that runs out of results replays the previous window** instead of returning an empty page, so each query deduplicates by `link` and stops once a page contributes nothing new.
3. **`num` is an approximate cap, not an exact count**, so surplus rows are trimmed client side before they reach the result limit.

Refusals arrive as a `detail` string in place of organic rows and are reported as `provider-error`. Everything else routes through `provider_http_error`, matching the other keyed sources. Paging is bounded at ten pages per query, which is a provider boundary guard rather than a result cap.

Emails are routed opportunistically, exactly as Brave routes them: SERP descriptions carry hostnames reliably and addresses rarely. I am not overselling that route.

## Test plan

- [x] `uv run pytest` -> 2293 passed, 6 skipped, 41 deselected
- [x] `uv run ruff check .` -> All checks passed
- [x] `uv run ruff format --check .` -> 154 files already formatted
- [x] `uv run ty check` -> All checks passed
- [x] `tests/discovery/test_serply.py` carries `pytest.mark.provider_contract('serply')` and needs no key and no network

## Real run

```
$ uv run theHarvester -d mozilla.org -b serply -l 20
[*] Hosts found: 10
blog.mozilla.org      builders.mozilla.org   connect.mozilla.org
donate.mozilla.org    foundation.mozilla.org hacks.mozilla.org
monitor.mozilla.org   newproducts.mozilla.org stateof.mozilla.org
www.mozilla.org

$ uv run theHarvester -d mozilla.org -b serply -l 50
[*] Hosts found: 15
(adds addons, blog.nightly, community, discourse, people)
```

The jump from 10 to 15 is the pagination check: `start` genuinely advances the window rather than replaying page one.

With no key configured:

```
[!] Source serply skipped: missing credentials.
```

Based on `dev` per AGENTS.md. Provider docs: https://serply.io/docs


NotoriousRebel:
Thanks for the PR, I will take a look

NotoriousRebel:
Review from Codex for `ab3a5f96`.

Two repository standards issues:

- P2: Add assertions for session reuse and cleanup to the [provider contract](https://github.com/laramies/theHarvester/blob/ab3a5f96e2efeddcc80153fefdaae6f9ca5cbbc5/tests/discovery/test_serply.py#L84-L100). The fetch mocks ignore the supplied session, so the tests don't prove that later pages retain earlier session state or that the session closes after success and cancellation. The [module guide](https://github.com/laramies/theHarvester/blob/ab3a5f96e2efeddcc80153fefdaae6f9ca5cbbc5/docs/wiki/How-to-add-a-new-module.md#L40-L51) requires these checks. Additional offline probes passed with the current implementation; the submitted tests need that coverage.
- P3: Remove the fixed "one API credit" cost from [README.md](https://github.com/laramies/theHarvester/blob/ab3a5f96e2efeddcc80153fefdaae6f9ca5cbbc5/README.md#L321) and the [configuration guide](https://github.com/laramies/theHarvester/blob/ab3a5f96e2efeddcc80153fefdaae6f9ca5cbbc5/docs/wiki/Configuration-and-API-Keys.md#L54). The [contribution guidance](https://github.com/laramies/theHarvester/blob/ab3a5f96e2efeddcc80153fefdaae6f9ca5cbbc5/docs/wiki/How-to-add-a-new-module.md#L14) asks contributors to link provider pricing and quotas rather than hardcode them. Keep the explanation of request bounds and link the current provider terms.

One spec issue:

- P2: Deduplicate the full response before trimming to the remaining result allowance. At [serplysearch.py:109](https://github.com/laramies/theHarvester/blob/ab3a5f96e2efeddcc80153fefdaae6f9ca5cbbc5/theHarvester/discovery/serplysearch.py#L109-L135), `rows[:remaining]` runs before link deduplication. With `limit=15`, a first page containing links 0 through 9 and a second page containing links 5 through 14, the slice keeps only duplicates 5 through 9. The adapter then ends that query and drops new links 10 through 14, even though the provider already returned them. This conflicts with the PR's stated rule to stop when a page contributes nothing new. Deduplicate the full page, trim the fresh rows, and use the full response to determine repetition or a short page.

Both overlap cases reproduced offline. The first page contained links 0 through 9, and the subsequent `site:` query returned an empty list:

| Second-page links | Requested limit | Expected retained rows | Actual retained rows |
| --- | --- | --- | --- |
| 5 through 14 | 15 | 15 | 10 |
| 8 through 17 | 15 | 15 | 13 |

Both returned `None`, which reports normal completion. An overlapping final-page regression would catch this.

Validation: 260 existing focused tests passed, along with Ruff lint/format checks for the changed production Python files and the diff whitespace check. The two added pagination assertions failed as described above; two added session lifecycle probes passed. Tests reused the existing environment. The full suite and authenticated live API behavior were not checked.

There is also a documentation discrepancy to clarify: Serply's [pagination guide](https://serply.io/docs/guides/pagination) describes page/cursor pagination, while the PR reports that only `start` advances this endpoint. That discrepancy is separate from the reproduced result-loss bug.

googio:
Thanks for the Codex pass. All four points are addressed in f0173ab.

**Dedup before trim (P2 spec).** The page is now deduplicated in full and only the fresh rows are trimmed to the remaining allowance, so a second page of 5-14 or 8-17 after 0-9 with `-l 15` retains 15 rows. The repeat-window stop and the short-page stop are judged on the full response. `test_serply_keeps_fresh_rows_from_an_overlapping_final_page` covers both offsets and fails on `ab3a5f96`, retaining 10 and 13 rows as in your table.

**Session reuse and cleanup (P2 standards).** The evidence test now records the session handed to every `fetch_json` call and asserts both pages borrow the one session opened by `open_session`, and that it closes after success. The cancellation test asserts the session closes when the fetch is cancelled. The overlap regression also asserts the second page reuses the first page's session.

**Docs (P3).** The fixed "one API credit" cost is gone from README.md and Configuration-and-API-Keys.md. Both keep the ten-rows-per-request bound and link https://serply.io/pricing for current terms.

**Pagination guide discrepancy.** The guide at https://serply.io/docs/guides/pagination now reads: "Serply passes Google's own pagination parameters straight through, so paging is offset-based: you ask for a page size with num and a starting offset with start. There are no cursors and no pagination envelope in the response." That matches what this module does. I also probed the live endpoint: `page=2` and `page=3` return windows that mostly overlap `start=0`, `start=10` and `start=20` move the window, and no response carries a page, cursor or next field.

Validation: the 30 tests in `tests/discovery/test_serply.py` pass and `ruff format --check` is clean on both changed Python files. `ruff check` with the local 0.16.3 reports TC003 on the pre-existing `AsyncIterator` import, which `test_brave.py` shares, so I left that as it was.

